PR #22, opened 2017-10-28
Updated RT goal handle to handle cancel requests

nlamprian:
With the proposed changes, the real time goal handle allows for a cancel request to happen and the action to be terminated at a later time from within the real time thread.

### DIFF
--- a/include/realtime_tools/realtime_server_goal_handle.h
+++ b/include/realtime_tools/realtime_server_goal_handle.h
@@ -55,6 +55,7 @@ private:
   uint8_t state_;
 
   bool req_abort_;
+  bool req_cancel_;
   bool req_succeed_;
   ResultConstPtr req_result_;
   FeedbackConstPtr req_feedback_;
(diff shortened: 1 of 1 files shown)


include/realtime_tools/realtime_server_goal_handle.h
```diff
@@ -55,6 +55,7 @@ class RealtimeServerGoalHandle
   uint8_t state_;
 
   bool req_abort_;
+  bool req_cancel_;
```
ipa-mdl: needs to be initialized in constructor

ipa-mdl:
This might help to fix some problems with canceling in `joint_trajectory_controller`, perhaps even https://github.com/ros-controls/ros_controllers/issues/48

bmagyar:
I'm not sure what is the best way to force travis to re-parse the CI configuration... Could you please rebase the PR on the latest `kinetic-devel`?

bmagyar:
Great, thank you! :+1:

bmagyar:
Should we cherry-pick this back into Indigo?

PR #25, opened 2018-04-27
Fix NON_POLLING version of realtime_publisher

kjaget:
As is, the code didn't compile with NON_POLLING defined.  Here's a pull request which fixes that.  It modifies lock and unlock for the NON_POLLING case to use a boost::unique_lock to lock rather than a bare mutex. This was necessary since that's what is used by boost::condition_variable wait and notify.


bmagyar:
Hi,

Any chances of reopening this against melodic-devel with C++11/14 features?

PR #26, opened 2018-05-03
C++11 and melodic travis updates

bmagyar:
`RealtimePublisherSharedPtr` is to be used in `ros_controllers` to replace some looong `boost::shared_ptr` declarations and typedef..

### DIFF
--- a/include/realtime_tools/realtime_publisher.h
+++ b/include/realtime_tools/realtime_publisher.h
@@ -227,6 +227,10 @@ private:
   int turn_;  // Who's turn is it to use msg_?
 };
 
+#include <memory>
+template <class Msg>
+using RealtimePublisherSharedPtr = std::shared_ptr<RealtimePublisher<Msg> >;
+
 }
 
 #endif
--- a/.travis.yml
+++ b/.travis.yml
@@ -14,14 +14,11 @@ notifications:
 
 env:
   global:
-    - UPSTREAM_WORKSPACE=https://raw.github.com/ros-controls/ros_control/kinetic-devel/ros_control.rosinstall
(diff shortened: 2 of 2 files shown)


bmagyar:
RealtimePublisherSharedPtr is to be used in `ros_controllers` to replace some looong `boost::shared_ptr` declarations and typedef..

include/realtime_tools/realtime_publisher.h
```diff
@@ -227,6 +227,10 @@ class RealtimePublisher : boost::noncopyable
   int turn_;  // Who's turn is it to use msg_?
 };
 
+#include <memory>
+template <class Msg>
+using RealtimePublisherSharedPtr = std::shared_ptr<RealtimePublisher<Msg> >;
+
```
ipa-mdl: This could be moved into the class as well.
Not sure which style to follow here..

bmagyar: Here, please correct if I'm wrong but I think this is how the alternatives would look like:

Eg:
`boost::shared_ptr<realtime_tools::RealtimePublisher<sensor_msgs::JointState> > realtime_pub_;`
to 
`realtime_tools::RealtimePublisher<sensor_msgs::JointState>::SharedPtr realtime_pub_;`
or
`realtime_tools::RealtimePublisherSharedPtr<sensor_msgs::JointState> realtime_pub_;`

I'd personally opt for less namespacing.

ipa-mdl: I was just curious if there are any good reason to prefer one over the other.

> I'd personally opt for less namespacing.

:+1:

PR #27, opened 2018-06-20
remove duplicated action pointer types  from RealtimeServerGoalHandle

ipa-mdl:
addresses https://github.com/ros-controls/ros_controllers/issues/352


ipa-mdl:
The might be needed for older distros as well.

ipa-mdl:
As an alternative we could remove the `ACTION_DEFINITION` call and add the other pointer types manually.
This would allow us to use `std::shared_ptr` instead of `boost::shared_ptr`.

PR #28, opened 2018-08-15
Made RealtimeBuffer's copy-constructor const

Matthew-Reynolds:
The copy constructor on the realtime buffer doesn't follow the standard copy-constructor format of pass-by-const-reference, but rather uses pass-by-reference. There is no reason for this, it should be const to be const-correct and follow standard copy-constructor idioms. This change also will allow RealtimeBuffer to be an STL container element.

I've opened this PR against `melodic-devel`, but the file `realtime_buffer.h` is identical between kinetic and melodic so ideally I would like to see this change cherry picked backward to at least `kinetic-devel`.

### DIFF
--- a/include/realtime_tools/realtime_buffer.h
+++ b/include/realtime_tools/realtime_buffer.h
@@ -76,7 +76,7 @@ class RealtimeBuffer
       delete realtime_data_;
   }
 
-  RealtimeBuffer(RealtimeBuffer &source)
+  RealtimeBuffer(const RealtimeBuffer &source)
   {
     // allocate memory
     non_realtime_data_ = new T();


bmagyar:
I think this is safe to do, may not fully be ABI compatible though...

Since this is a tad bit more of a delicate matter, I'll summon:
@ipa-mdl @graiola and @efernandez 

Note, for comparison see:
`RealtimeBuffer &operator =(const RealtimeBuffer& source)`

ipa-mdl:
> I think this is safe to do, may not fully be ABI compatible though...

it's header only, so the ABI should be the safe.

efernandez:
:+1:

Matthew-Reynolds:
I am unfamiliar with conventions for maintaining previous but still supported branches, but I was hoping to discuss having this change cherry picked backwards to Kinetic given Kinetic is still supported (And is still the recommended release) and the changes are minimal & easy to cherry pick. What are the conventions for changes to older releases? Looking to learn. Thanks.

matthew-reynolds:
Can we cherry pick this back to Kinetic? As mentioned, it's a header-only lib so there shouldn't be ABI concerns

PR #31, opened 2018-12-09
Fix for NON_POLLING realtime_publisher

kjaget:
Sorry for opening a new PR - couldn't see a way to change the source branch from kinetic to melodic.  So this is an update of PR 25 to melodic-devel

As is, the code didn't compile with NON_POLLING defined. Here's a pull request which fixes that. It modifies lock and unlock for the NON_POLLING case to use a boost::unique_lock to lock rather than a bare mutex. This was necessary since that's what is used by boost::condition_variable wait and notify.

In PR 25, it was mentioned to update this to use c++11 / 14 features. Does that mean move to std:: instead of boost::?

### DIFF
--- a/include/realtime_tools/realtime_publisher.h
+++ b/include/realtime_tools/realtime_publisher.h
@@ -179,7 +179,6 @@ private:
     thread_ = boost::thread(&RealtimePublisher::publishingLoop, this);
   }
 
-
   bool is_running() const { return is_running_; }
 
   void publishingLoop()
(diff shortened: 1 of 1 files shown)


bmagyar:
Sorry for the delay.

Yes, if it's not a problem, please use `std::` instead of `boost::`

kjaget:
My turn to be sorry for the delay.  The latest update I just pushed is a rebase to line up with the latest melodic-devel along with the requested changes to move away from boost:: to the corresponding objects in std::.

bmagyar:
I believe https://github.com/ros-controls/realtime_tools/pull/39 brought these changes in, I'm closing now but feel free to reopen if you don't agree.

PR #32, opened 2019-01-31
use this_thread::sleep_for on Windows

kejxu:
`usleep` is not available on Windows, replace it with `this_thread::sleep_for` (available since c++11)

also a minor indentation update in include/realtime_tools/realtime_publisher.h, converted existing tabs to spaces


kejxu:
> Why did you keep `usleep`? For melodic we can switch to `std::chrono` right away.

we are keeping `usleep` just so it would not change the existing behavior on Linux, will remove the macro guard and apply this change to both Linux and Windows

PR #34, opened 2019-03-01
add comment for install command

kejxu:
http://docs.ros.org/melodic/api/catkin/html/howto/format2/building_libraries.html


kejxu:
@ipa-mdl, only adding comment, could we get a quick merge for this one? =)

btw, sorry for the spam!

ipa-mdl:
Actually, I don't see the need for this comment..
None of the other sections is commented and it is kinda obvious what's happening in these lines.

bmagyar:
Agree. No need for commenting it

kejxu:
> Actually, I don't see the need for this comment..
> None of the other sections is commented and it is kinda obvious what's happening in these lines.

On a second thought, I have to agree with @ipa-mdl and @bmagyar.

Thanks for the prompt response, closing the pull request =)

Issue #35, opened 2019-05-22
Porting to ROS 2 Dashing


sloretz:
I'm looking into how much work it would be to port this package into ROS 2 Dashing. Mind answering a few questions?

* Is someone already working on this?
* Where should a ros2 port live? How about a `dashing-devel` branch in this repo?
* Would the maintainers be willing to release into Dashing once code has been ported?
* Should the port also eliminate the use of boost? https://github.com/ros-controls/realtime_tools/pull/31#issuecomment-452314598
* Is a partial port OK?
  * For example, I would like to use `RealtimePublisher`, but I don't currently need `RealtimeServerGoalHandle`. Would porting these classes one at a time be acceptable?
* Should the existing code style be maintained, or should the [ROS 2 style guide](https://index.ros.org/doc/ros2/Contributing/Developer-Guide/#id1) be used?

bmagyar:
Hi @sloretz , 
Great list, allow me to answer in the same fashion.

* Not that I know of.
* Works perfectly fine.
* No problem.
* As much as possible, I've been trying to get rid of most boost in favour of standard library.
* I think a partial port may be misleading and harder to maintain. This repo doesn't have that much code TBH. It may be a bigger issue elsewhere.
* Feel free to restyle things, but please try doing it in a separate PR from functional ones. I have only one problem with the style guide: I find `CamelCase` for functions terrible, also most of our existing API uses `camelCase` which is just way more readable and avoids confusing functions with classes.

Hope this helps!

ipa-mdl:
> I find CamelCase for functions terrible, also most of our existing API uses camelCase which is just way more readable and avoids confusing functions with classes.

Urgs, I did not see this before. :+1:  for `camelCase`.

sloretz:
@bmagyar now that #41 is merged :tada:, would you be willing to release realtime_tools into dashing?

bmagyar:
Sorry for the slight delay, here you go: https://github.com/ros/rosdistro/pull/22203

PR #37, opened 2019-05-23
Add unit tests + fix publisher bug

sloretz:
This PR adds unit tests to make sure stuff doesn't get broken while porting to ROS 2 (#35). Also included is a fix for a race condition where `RealtimePublisher` may not publish the first message. The unit test needs this fix to pass reliably.

`RealtimePublisher` initialized `turn_` with `REALTIME`, and `publishingLoop()` also sets `turn_` to `REALTIME`. If `unblockAndPublish()` is called before before the publishing thread gets scheduled then the first message won't be published because `turn_` is overwritten. This fixes the issue by initializing `turn_` to a new state `LOOP_NOT_STARTED`. This prevents `trylock()` from returning `true` because the state is not `REALTIME`, which means there is no message to get lost.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -29,6 +29,11 @@ if (TARGET realtime_buffer_tests)
   target_link_libraries(realtime_buffer_tests PRIVATE ${PROJECT_NAME})
 endif()
 
+catkin_add_gtest(realtime_clock_tests test/realtime_clock_tests.cpp)
+if (TARGET realtime_clock_tests)
+  target_link_libraries(realtime_clock_tests PRIVATE ${PROJECT_NAME})
+endif()
+
 # Install
 install(DIRECTORY include/${PROJECT_NAME}/
   DESTINATION ${CATKIN_PACKAGE_INCLUDE_DESTINATION})
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -42,6 +42,13 @@ if (CATKIN_ENABLE_TESTING)
     target_link_libraries(realtime_publisher_tests ${PROJECT_NAME} ${std_msgs_LIBRARIES})
     target_include_directories(realtime_publisher_tests PRIVATE ${std_msgs_INCLUDE_DIRS})
   endif()
+
+  add_rostest_gtest(realtime_server_goal_handle_tests test/realtime_server_goal_handle.test test/realtime_server_goal_handle_tests.cpp)
+  if (TARGET realtime_server_goal_handle_tests)
+    find_package(actionlib REQUIRED)
+    target_link_libraries(realtime_server_goal_handle_tests ${PROJECT_NAME} ${actionlib_LIBRARIES})
+    target_include_directories(realtime_publisher_tests PRIVATE ${actionlib_INCLUDE_DIRS})
+  endif()
 endif()
 
 # Install
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -21,34 +21,24 @@ target_link_libraries(${PROJECT_NAME} PUBLIC ${catkin_LIBRARIES} ${Boost_LIBRARI
 # Unit Tests
 if (CATKIN_ENABLE_TESTING)
   catkin_add_gtest(realtime_box_tests test/realtime_box_tests.cpp)
-  if (TARGET realtime_box_tests)
-    target_link_libraries(realtime_box_tests ${PROJECT_NAME})
-  endif()
+  target_link_libraries(realtime_box_tests ${PROJECT_NAME})
 
   catkin_add_gtest(realtime_buffer_tests test/realtime_buffer_tests.cpp)
-  if (TARGET realtime_buffer_tests)
-    target_link_libraries(realtime_buffer_tests ${PROJECT_NAME})
-  endif()
+  target_link_libraries(realtime_buffer_tests ${PROJECT_NAME})
 
   catkin_add_gtest(realtime_clock_tests test/realtime_clock_tests.cpp)
-  if (TARGET realtime_clock_tests)
-    target_link_libraries(realtime_clock_tests ${PROJECT_NAME})
-  endif()
+  target_link_libraries(realtime_clock_tests ${PROJECT_NAME})
 
   find_package(rostest REQUIRED)
   add_rostest_gtest(realtime_publisher_tests test/realtime_publisher.test test/realtime_publisher_tests.cpp)
-  if (TARGET realtime_publisher_tests)
-    find_package(std_msgs REQUIRED)
-    target_link_libraries(realtime_publisher_tests ${PROJECT_NAME} ${std_msgs_LIBRARIES})
-    target_include_directories(realtime_publisher_tests PRIVATE ${std_msgs_INCLUDE_DIRS})
-  endif()
+  find_package(std_msgs REQUIRED)
+  target_link_libraries(realtime_publisher_tests ${PROJECT_NAME} ${std_msgs_LIBRARIES})
+  target_include_directories(realtime_publisher_tests PRIVATE ${std_msgs_INCLUDE_DIRS})
 
   add_rostest_gtest(realtime_server_goal_handle_tests test/realtime_server_goal_handle.test test/realtime_server_goal_handle_tests.cpp)
-  if (TARGET realtime_server_goal_handle_tests)
-    find_package(actionlib REQUIRED)
-    target_link_libraries(realtime_server_goal_handle_tests ${PROJECT_NAME} ${actionlib_LIBRARIES})
-    target_include_directories(realtime_publisher_tests PRIVATE ${actionlib_INCLUDE_DIRS})
-  endif()
+  find_package(actionlib REQUIRED)
+  target_link_libraries(realtime_server_goal_handle_tests ${PROJECT_NAME} ${actionlib_LIBRARIES})
+  target_include_directories(realtime_publisher_tests PRIVATE ${actionlib_INCLUDE_DIRS})
 endif()
 
 # Install
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -20,22 +20,22 @@ target_link_libraries(${PROJECT_NAME} PUBLIC ${catkin_LIBRARIES} ${Boost_LIBRARI
 
 # Unit Tests
 if (CATKIN_ENABLE_TESTING)
-  catkin_add_gtest(realtime_box_tests test/realtime_box_tests.cpp)
-  target_link_libraries(realtime_box_tests ${PROJECT_NAME})
+  catkin_add_gmock(realtime_box_tests test/realtime_box_tests.cpp)
+  target_link_libraries(realtime_box_tests ${PROJECT_NAME} ${GMOCK_MAIN_LIBRARIES})
 
-  catkin_add_gtest(realtime_buffer_tests test/realtime_buffer_tests.cpp)
-  target_link_libraries(realtime_buffer_tests ${PROJECT_NAME})
+  catkin_add_gmock(realtime_buffer_tests test/realtime_buffer_tests.cpp)
+  target_link_libraries(realtime_buffer_tests ${PROJECT_NAME} ${GMOCK_MAIN_LIBRARIES})
 
-  catkin_add_gtest(realtime_clock_tests test/realtime_clock_tests.cpp)
+  catkin_add_gmock(realtime_clock_tests test/realtime_clock_tests.cpp)
   target_link_libraries(realtime_clock_tests ${PROJECT_NAME})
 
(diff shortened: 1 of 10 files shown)


CMakeLists.txt
```diff
@@ -16,7 +16,40 @@ catkin_package(
   )
 
 add_library(${PROJECT_NAME} src/realtime_clock.cpp)
-target_link_libraries(${PROJECT_NAME} ${catkin_LIBRARIES} ${Boost_LIBRARIES})
+target_link_libraries(${PROJECT_NAME} PUBLIC ${catkin_LIBRARIES} ${Boost_LIBRARIES})
+
+# Unit Tests
+if (CATKIN_ENABLE_TESTING)
+  catkin_add_gtest(realtime_box_tests test/realtime_box_tests.cpp)
```
bmagyar: How do you feel about using GMOCK instead? The majority wouldn't change now but I think that moving forward it'd be valuable.

CMakeLists.txt
```diff
@@ -16,7 +16,40 @@ catkin_package(
...
+  if (TARGET realtime_box_tests)
```
bmagyar: Please forgive my ignorance on this one but I forgot the why of this hacky IF setup. Can we avoid using it?

test/realtime_box_tests.cpp
```diff
@@ -0,0 +1,79 @@
...
+    const char input = 'z';
+    box.set(input);
+  }
+
+  char output = 'a';
+  box.get(output);
+  EXPECT_EQ('z', output);
+}
+
+int main(int argc, char ** argv) {
```
bmagyar: This standard main could be removed and replaced by linking to `GMOCK_MAIN_LIBRARIES` in the cmake. I think it's always better when we have less C++ to maintain, no matter how simple.

test/realtime_publisher_tests.cpp
```diff
@@ -0,0 +1,99 @@
...
+  const size_t ATTEMPTS = 10;
+  const std::chrono::milliseconds DELAY(250);
+
+  const char * expected_msg = "Hello World";
+  ros::NodeHandle nh;
+  const bool latching = true;
+  RealtimePublisher<std_msgs::String> rt_pub(nh, "rt_publish", 10, latching);
+  // publish a latched message
+  bool lock_is_held = rt_pub.trylock();
+  for (size_t i = 0; i < ATTEMPTS && !lock_is_held; ++i) {
```
bmagyar: please drop the `{` to a new line

test/realtime_clock_tests.cpp
```diff
@@ -0,0 +1,59 @@
...
+
+TEST(RealtimeClock, get_system_time)
+{
+  const int ATTEMPTS = 10;
+  const std::chrono::milliseconds DELAY(1);
+
+  RealtimeClock rt_clock;
+  // Wait for time to be available
+  ros::Time last_rt_time;
+  for (int i = 0; i < ATTEMPTS && ros::Time() == last_rt_time; ++i) {
```
bmagyar: please drop the `{` to a new line

test/realtime_server_goal_handle_tests.cpp
```diff
@@ -0,0 +1,280 @@
...
+    have_handle_ = true;
+  }
+
+  void cancel_callback(GoalHandle handle)
+  {
+  }
+
+  bool wait_for_handle()
+  {
+    for (size_t i = 0; i < ATTEMPTS; ++i) {
```
bmagyar: please drop the `{` to a new line

test/realtime_server_goal_handle_tests.cpp
```diff
@@ -0,0 +1,280 @@
...
+      ros::spinOnce();
+      std::this_thread::sleep_for(DELAY);
+      std::unique_lock<std::mutex> lock(mtx_);
+      if (have_handle_) {
```
bmagyar: please drop the `{` to a new line

test/realtime_server_goal_handle_tests.cpp
```diff
@@ -0,0 +1,280 @@
...
+  using FeedbackConstPtr = actionlib::ActionServer<actionlib::TwoIntsAction>::FeedbackConstPtr;
+  void feedback_callback(const FeedbackConstPtr &)
+  {
+    std::unique_lock<std::mutex> lock(mtx_);
+    have_feedback_ = true;
+  }
+
+  bool wait_for_feedback()
+  {
+    for (size_t i = 0; i < ATTEMPTS; ++i) {
```
bmagyar: please drop the `{` to a new line

test/realtime_server_goal_handle_tests.cpp
```diff
@@ -0,0 +1,280 @@
...
+      ros::spinOnce();
+      std::this_thread::sleep_for(DELAY);
+      std::unique_lock<std::mutex> lock(mtx_);
+      if (have_feedback_) {
```
bmagyar: please drop the `{` to a new line

test/realtime_server_goal_handle_tests.cpp
```diff
@@ -0,0 +1,280 @@
...
+
+std::shared_ptr<actionlib::SimpleActionClient<actionlib::TwoIntsAction>>
+send_goal(
+  const std::string & server_name,
+  FeedbackCallback * cb = nullptr)
+{
+  std::shared_ptr<actionlib::SimpleActionClient<actionlib::TwoIntsAction>> ac;
+  ac.reset(new actionlib::SimpleActionClient<actionlib::TwoIntsAction>(server_name, true));
+
+  for (size_t i = 0; i < ATTEMPTS && !ac->isServerConnected(); ++i) {
```
bmagyar: please drop the `{` to a new line

test/realtime_server_goal_handle_tests.cpp
```diff
@@ -0,0 +1,280 @@
...
+    ros::spinOnce();
+    std::this_thread::sleep_for(DELAY);
+  }
+  if (!ac->isServerConnected()) {
```
bmagyar: please drop the `{` to a new line

test/realtime_server_goal_handle_tests.cpp
```diff
@@ -0,0 +1,280 @@
...
+      if (have_feedback_) {
+        break;
+      }
+    }
+    std::unique_lock<std::mutex> lock(mtx_);
+    return have_feedback_;
+  }
+};
+
+std::shared_ptr<actionlib::SimpleActionClient<actionlib::TwoIntsAction>>
```
bmagyar: Could we do something like this please? 
```suggestion
using TwoIntsActionClient = actionlib::SimpleActionClient<actionlib::TwoIntsAction>;
std::shared_ptr<TwoIntsActionClient>
```

test/realtime_server_goal_handle_tests.cpp
```diff
@@ -0,0 +1,280 @@
...
+send_goal(
+  const std::string & server_name,
+  FeedbackCallback * cb = nullptr)
+{
+  std::shared_ptr<actionlib::SimpleActionClient<actionlib::TwoIntsAction>> ac;
```
bmagyar: Let's get [crazy](https://www.youtube.com/watch?v=u9sJ4NAugxg)!
```suggestion
  auto ac = std::make_shared<TwoIntsActionClient>(server_name, true);
```

and with this the next line can be dropped. (Looks fine to me but if it doesn't compile, you get a :cookie:  from Karsten)

test/realtime_box_tests.cpp
bmagyar: Of course this doesn't apply in cases where you have to call `ros::init`.

bmagyar:
@kjaget could you please confirm whether this addresses #31  too?

CMakeLists.txt
sloretz: I was copying [an example from the ros wiki](http://wiki.ros.org/gtest#Update_CMakeLists.txt). It looks like `catkin_add_gtest()` [always calls `add_executable()`](https://github.com/ros/catkin/blob/94c7d2e25ee2aa6b1f19a41f1deda731c7f8af63/cmake/test/gtest.cmake#L171) so this check shouldn't be necessary.

sloretz: It looks like rostest [only creates a target if gtest/gmock is found](https://github.com/ros/ros_comm/blob/29053c4832229efa7160fb944c05e3bc82e11540/tools/rostest/cmake/rostest-extras.cmake.em#L78).

CMakeLists.txt
sloretz: Switched to gmock in ac9bf90f27c2c8464c9c7cf2775960ca1ba3d95f

CMakeLists.txt
sloretz: Removed all `if(TARGET ...)` checks in be439e97931e55d59d8500f310046f41572a1a54. If `gtest/gmock` isn't found then this code will complain at configure time.

test/realtime_box_tests.cpp
sloretz: Uses `GMOCK_MAIN_LIBRARIES` for box and buffer tests in ac9bf90f27c2c8464c9c7cf2775960ca1ba3d95f

test/realtime_publisher_tests.cpp
sloretz: `{` on new lines in 8d81fc5e56aaae72e07b768df9d68b334b3bdd0c

test/realtime_server_goal_handle_tests.cpp
sloretz: Shorter type names for action client and server in 709ce67dbb68ec6e9baed20901ee5abbcc3bdfee

test/realtime_server_goal_handle_tests.cpp
sloretz: It compiled, no :cookie: for me :(

Use `std::make_shared` for server and client in 717f5473a782a9d04ef55550dc9afc331536ffa9

test/realtime_clock_tests.cpp
sloretz: 8d81fc5e56aaae72e07b768df9d68b334b3bdd0c

test/realtime_server_goal_handle_tests.cpp
sloretz: 8d81fc5e56aaae72e07b768df9d68b334b3bdd0c

test/realtime_server_goal_handle_tests.cpp
sloretz: 8d81fc5e56aaae72e07b768df9d68b334b3bdd0c

test/realtime_server_goal_handle_tests.cpp
sloretz: 8d81fc5e56aaae72e07b768df9d68b334b3bdd0c

test/realtime_server_goal_handle_tests.cpp
sloretz: 8d81fc5e56aaae72e07b768df9d68b334b3bdd0c

test/realtime_server_goal_handle_tests.cpp
sloretz: 8d81fc5e56aaae72e07b768df9d68b334b3bdd0c

test/realtime_server_goal_handle_tests.cpp
sloretz: 8d81fc5e56aaae72e07b768df9d68b334b3bdd0c

sloretz:
@bmagyar FYI I believe all the comments from your review have been addressed, so this is ready for another review.

CMakeLists.txt
bmagyar: Thanks. I actually had an unsubmitted comment on this:

"Yes. And in the case they aren't, I think the compilation should stop, error out and potentially set the computer on fire. If there are tests and we set the flag to compile tests, they should be compiled."

test/realtime_server_goal_handle_tests.cpp
bmagyar: Thanks!

CMakeLists.txt
bmagyar: Thank you

CMakeLists.txt
bmagyar: Great!

test/realtime_server_goal_handle_tests.cpp
bmagyar: :+1:

bmagyar:
@kjaget and @ipa-mdl if you guys could give this PR a quick read I'll buy you a beer the next time we meet.

PR #38, opened 2019-05-24
Package format 2

sloretz:
~~Please review #37 first. I'll rebase this once #37 is through the review processes. The only file changed by this PR is `package.xml`~~ Edit: Changed my mind, it looks like this can go in without creating a merge conflict with #37 so this PR is ready to be reviewed

This PR
* Updates the `package.xml` to use format 2
* Adds a `<build_export_depend>` for `actionlib` since its headers are included by headers installed by this package
* Shortens the package.xml description since it appeared to be out of date

This is a step towards #35

### DIFF
--- a/package.xml
+++ b/package.xml
@@ -17,7 +17,6 @@
   <buildtool_depend>catkin</buildtool_depend>
 
   <depend>roscpp</depend>
-  <depend>rospy</depend>
 
   <build_export_depend>actionlib</build_export_depend>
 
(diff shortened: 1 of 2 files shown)


package.xml
```diff
@@ -24,10 +16,9 @@
 
   <buildtool_depend>catkin</buildtool_depend>
 
-  <build_depend>roscpp</build_depend> 
-  <build_depend>rospy</build_depend> 
+  <depend>roscpp</depend>
+  <depend>rospy</depend>
```
bmagyar: With the script gone, we can actually say goodbye to this dependency too ;)

sloretz: Rebased and removed rclpy dependency in 460092b

PR #39, opened 2019-05-24
Remove boost

sloretz:
This removes all uses of boost. This PR depends on #37. I'll rebase this one once that's through the review process.

This change probably breaks ABI (changes the size of `RealtimeClock`), so it should be targeted at a `noetic-devel` branch.

This is a step towards #35

### DIFF
--- a/include/realtime_tools/realtime_server_goal_handle.h
+++ b/include/realtime_tools/realtime_server_goal_handle.h
@@ -46,8 +46,8 @@ private:
   ACTION_DEFINITION(Action);
 
   typedef actionlib::ServerGoalHandle<Action> GoalHandle;
-  typedef boost::shared_ptr<Result> ResultPtr;
-  typedef boost::shared_ptr<Feedback> FeedbackPtr;
+  typedef std::shared_ptr<Result> ResultPtr;
+  typedef std::shared_ptr<Feedback> FeedbackPtr;
 
   uint8_t state_;
 
--- a/include/realtime_tools/realtime_server_goal_handle.h
+++ b/include/realtime_tools/realtime_server_goal_handle.h
@@ -33,6 +33,8 @@
 // Standard
 #include <memory>
 
+#include <boost/shared_ptr.hpp>
+
 // Actionlib
 #include <actionlib/server/action_server.h>
 
@@ -46,8 +48,8 @@ private:
   ACTION_DEFINITION(Action);
 
   typedef actionlib::ServerGoalHandle<Action> GoalHandle;
-  typedef std::shared_ptr<Result> ResultPtr;
-  typedef std::shared_ptr<Feedback> FeedbackPtr;
+  typedef boost::shared_ptr<Result> ResultPtr;
+  typedef boost::shared_ptr<Feedback> FeedbackPtr;
 
   uint8_t state_;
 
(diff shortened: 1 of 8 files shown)


include/realtime_tools/realtime_server_goal_handle.h
```diff
@@ -46,8 +46,8 @@ class RealtimeServerGoalHandle
   ACTION_DEFINITION(Action);
 
   typedef actionlib::ServerGoalHandle<Action> GoalHandle;
-  typedef boost::shared_ptr<Result> ResultPtr;
-  typedef boost::shared_ptr<Feedback> FeedbackPtr;
+  typedef std::shared_ptr<Result> ResultPtr;
```
bmagyar: This may be a little bit of a slippery slope. I suggest you guys coordinate with @clalancette on this

https://github.com/ros-controls/ros_controllers/issues/352

Long story short: we shouldn't even be hosting these typedefs here as they should come from actionlib where boost is still used.

sloretz: If I read up on the story correctly, it looks like this can be fixed in `melodic`. https://github.com/ros-controls/ros_controllers/issues/352#issuecomment-496619360

Afterwards, I should be able to delete the typedefs from this PR

bmagyar:
Could you please do an interactive rebase and drop commit of #37 from this branch? I suppose we got the conflict from the changes applied to that one during the review.

sloretz:
@bmagyar Rebased, however I haven't yet rebuilt this locally after rebasing.

Also, do you care if the commit history is organized? I've gotten in the habit of not cleaning up my commits because we usually squash and merge PRs on the ros2 org, but I notice the unit test PR was rebased onto the target branch instead of squashed.

include/realtime_tools/realtime_server_goal_handle.h
bmagyar: I'd opt for keeping this one on a separate MR if you don't mind. This makes the package unreleasable until the above one is solved. I'd like to push out a release of all these changes as soon as we are through the reviews.

src/realtime_clock.cpp
```diff
@@ -49,10 +48,9 @@ namespace realtime_tools
      system_time_(ros::Time()), 
      last_realtime_time_(ros::Time()),
      running_(true),
-     initialized_(false)
+     initialized_(false),
```
bmagyar: Do you want to keep this style of initialisation or should we adopt the newer trend of initialising them in the class definition? It'd limit the scope of this constructor to the thread setup.

bmagyar:
@sloretz only a few points pending on this one...

src/realtime_clock.cpp
sloretz: Moved initial values to class definition in 30109311ac8c8a88887f56f2ac20c08d1f47f0b0

include/realtime_tools/realtime_server_goal_handle.h
sloretz: Undid shared_ptr changes in 218a5f23bec267d25d5d38689fc07fe2dc344c6b

sloretz: Also, opened ros/actionlib#145

IanTheEngineer:
@bmagyar seems @sloretz addressed your final comments (though he didn't point it out :)

PR #43, opened 2019-08-15
Fix building realtime_publisher using NON_POLLING

kjaget:
An update from my previous PR since a lot has been improved since then. This should isolate it to just the fixes to get NON_POLLING to build.

Previously this tried wait on a condition variable using the lock() method as an argument. That didn't build because wait() expects a lock_guard<mutex> rather than a class method. This commit adds a lock_guard as a local var to the publisher loop and uses it to wait when NON_POLLING is defined. 

Additionally, it updates code to call the unlock() method in various places to pick up conditional calls to       the condition var's notify_one() to wake up the cv's wait().

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,25 @@
+# Travis Continuous Integration Configuration File For ROS Control Projects
+language: generic
+services:
+  - docker
+
+notifications:
+  email:
+    recipients:
+      - enrique.fernandez.perdomo@gmail.com
+      - bence.magyar.robotics@gmail.com
+      - gennaro.raiola@gmail.com
+    on_success: change #[always|never|change] # default: change
+    on_failure: change #[always|never|change] # default: always
+
+env:
+  global:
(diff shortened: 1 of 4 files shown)


destogl (approved):
This looks fine to me.

roncapat:
@bmagyar @destogl can this be merged?

.travis.yml
```diff
@@ -0,0 +1,25 @@
...
+language: generic
+services:
+  - docker
+
+notifications:
+  email:
+    recipients:
+      - enrique.fernandez.perdomo@gmail.com
+      - bence.magyar.robotics@gmail.com
+      - gennaro.raiola@gmail.com
```
bmagyar, line 11:
```suggestion
      - bence.magyar.robotics@gmail.com
```

bmagyar: does travis even work still? :D

bmagyar (approved):

PR #49, opened 2020-02-05
Check whether thread joinable before joining

Maverobot:
Current when `RealtimePublisher` is constructed with default constructor and `init` function is never called, joining an "empty" thread will cause exception. 

This PR checks whether the thread is joinable before joining.


bmagyar:
There has just been a melodic sync, I think there's at least 2weeks until
the next one. I'm pushing the release as we speak though.

Issue #50, opened 2020-02-09
thread-safe implementation in realtime_publisher


Maverobot:
When reading the code, I have a few questions regarding [realtime_publisher.h](https://github.com/ros-controls/realtime_tools/blob/melodic-devel/include/realtime_tools/realtime_publisher.h)

1. `volatile bool` is used for cross-threads flags. As far as I know, it is undefined behavior in C++. Should we replace that with `std::atomic<bool>`? 
    Some related discussions:
    * [C++ : std::atomic<bool> and volatile bool](https://stackoverflow.com/q/29633222/6018272)
    * [Is volatile bool for thread control considered wrong?](https://stackoverflow.com/questions/6995310/is-volatile-bool-for-thread-control-considered-wrong)
2. `int turn_` is also being used in both threads and not explicitly guarded by any mutex.
3. Why is `NO_POLLING` not activated by default? Does it mean the current release version is actually using polling instead of `condition_variable`, which is in my opinion a much better choice?

matthew-reynolds:
1. Indeed, `volatile bool` here is incorrect and dangerous. Looks like something written nearly 10 years ago that's been carried forward ever since. I don't think the solution is `atomic<bool>` though, I think `atomic_flag` is more suited since it is guaranteed to be implemented lock-free, while the `atomic<T>` types may be implemented with locks. (Practically, I imagine all implementation for an `atomic<bool>` are lock-free, but we should still strive for correctness.)
Looks like #51 was just opened to address this issue - Let's move this discussion there?

2. Agreed, `int turn_` also looks like a good candidate for an `std::atomic<>` type, this time `std::atomic<int>`.

3. I don't know why `NON_POLLING` isn't default (@bmagyar thoughts? My only guess is that when it was written in 2013, it required boost and was therefore "heavier"?), but from #43 it looks like it currently doesn't build anyways. Perhaps once it's fixed we should investigate enabling it by default, at least in Noetic.

bmagyar:
Thank you guys for chasing this.

I propose the following: 
* #43 to fix the state of affairs in Melodic
* #52 to bring a proper fix and to use the non-polling in Noetic
* #51 to spice up the ROS2 world, here I'd prefer to not disturb Eloquent but rather target Foxy. First and foremost, I'll update the master branch to host the latest ros2 branch and retarget #51 to that, then cut a foxy release

christophfroehlich:
Moving the last open point to #199

PR #51, opened 2020-02-14
Use std::atomic instead of volatile

youtalk:
C++11 has `std::atomic` for atomic variable. This PR uses it instead of classical `volatile` variable.

### DIFF
--- a/include/realtime_tools/realtime_publisher.h
+++ b/include/realtime_tools/realtime_publisher.h
@@ -205,8 +205,8 @@ private:
   }
 
   PublisherSharedPtr publisher_;
-  std::atomic_bool is_running_;
-  std::atomic_bool keep_running_;
+  std::atomic<bool> is_running_;
+  std::atomic<bool> keep_running_;
 
   std::thread thread_;
 
@@ -217,7 +217,7 @@ private:
 #endif
 
   enum {REALTIME, NON_REALTIME, LOOP_NOT_STARTED};
-  int turn_;  // Who's turn is it to use msg_?
+  std::atomic<int> turn_;  // Who's turn is it to use msg_?
 };
 
 template <class Msg>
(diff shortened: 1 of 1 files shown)


include/realtime_tools/realtime_publisher.h
```diff
@@ -204,8 +205,8 @@ class RealtimePublisher
   }
 
   PublisherSharedPtr publisher_;
-  volatile bool is_running_;
-  volatile bool keep_running_;
+  std::atomic_bool is_running_;
+  std::atomic_bool keep_running_;
```
matthew-reynolds: As mentioned in #50, `std::atomic_flag` is more suitable here since it's guaranteed to be implemented lock-free.

Maverobot: Problem with `std::atomic_flag` is that you cannot get its state without setting it to `true`. If we try to work around it then probably we will have to reinvent `std::atomic<bool>` :smile:

matthew-reynolds: Ah, I didn't know that - Thanks!

Maverobot: I figured that out when trying to implement your suggestion myself. :smiley:

youtalk:
Thank you for your reviews. Please consider to merge it.

bmagyar:
This should be an ABI breaking change. Let's be careful with them. Dashing is definitely off the radar here. Do you have other distro preferences!

Karsten1987:
I've just created a `ros2_devel` branch which is on pair with the current dashing branch. Feel free to re-target your PR here against `ros2_devel`. We can then cut a release for ROS2 Foxy from this.

matthew-reynolds:
I think this PR should be closed in favour of cherry-picking #52 to `ros2_devel`.

Karsten1987:
I'll leave this up to @youtalk 
I guess in any case these changes are very welcome on the ROS2 branches as well. So either way is fine.

youtalk:
I've changed the target branch. Please check again.

Karsten1987:
Looking at #52, what about the `atomic<int>`?

youtalk:
I've changed to use the `atomic<bool>` and `atomic<int>`.

PR #52, opened 2020-02-16
Use atomic types

Maverobot:
Use `std::atomic` types for safely accessing variables from different threads.

### DIFF
--- a/include/realtime_tools/realtime_publisher.h
+++ b/include/realtime_tools/realtime_publisher.h
@@ -55,7 +55,7 @@ class RealtimePublisher
 public:
   /// The msg_ variable contains the data that will get published on the ROS topic.
   Msg msg_;
-  
+
   /**  \brief Constructor for the realtime publisher
    *
    * \param node the nodehandle that specifies the namespace (or prefix) that is used to advertise the ROS topic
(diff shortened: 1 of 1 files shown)


bmagyar:
I suspect this change would introduce a binary incompatibility within all clients of realtime_tools and co.

Have you guys tested this?

We had a breaking change in melodic recently and would really like to avoid another one. Do you mind if we consider this for noetic instead?

matthew-reynolds:
@bmagyar `realtime_publisher` is header-only, so we should be safe from ABI breakages on our end, but indeed this may lead to issues in downstream binaries (ie classes that contain `RealtimePublisher` members.

From the standard: [atomics.types.generic]/p9:
>The representation of an atomic specialization need not have the same size as its corresponding
argument type. Specializations should have the same size whenever possible, as this reduces the effort to port existing code

Perhaps we should check the `sizeof(std::atomic<bool>)` and `sizeof(std::atomic<int>)` on GCC and Clang, and see if they were able to keep it the same size. Even if they were not able to, I believe we should still strongly consider this change for Melodic and even Kinetic, since we currently have dangerous, incorrect code.

bmagyar:
For now I have retargeted this PR to Noetic and going to push a release with it.
Let's bring the discussions from here over to #50 , I agree with your concerns however it seems that this dangerous piece of code took the test of time well so I'm a little hesitant about bringing these changes to earlier distros and breaking ABI in client packages.

Issue #44, opened 2019-09-09
unable to build on arch linux due to gmock not being found


jwhendy:
I posted [an issue](https://github.com/ros/catkin/issues/1030) at ros/catkin as well, since I first found issues related to gtest/gmock there. Downloading current from git and building via `catkin build` yields the following output for me:
```
Errors     << realtime_tools:cmake /home/jwhendy/catkin_ws/logs/realtime_tools/build.cmake.000.log                                                                                       
CMake Warning at /opt/ros/melodic/share/catkin/cmake/test/gtest.cmake:151 (message):
  skipping gmock 'realtime_box_tests' in project 'realtime_tools' because
  gmock was not found
Call Stack (most recent call first):
  /opt/ros/melodic/share/catkin/cmake/test/gtest.cmake:80 (_catkin_add_executable_with_google_test)
  /opt/ros/melodic/share/catkin/cmake/test/gtest.cmake:54 (_catkin_add_google_test)
  CMakeLists.txt:23 (catkin_add_gmock)


CMake Error at /home/jwhendy/catkin_ws/src/realtime_tools/CMakeLists.txt:24 (target_link_libraries):
  Cannot specify link libraries for target "realtime_box_tests" which is not
  built by this project.


cd /home/jwhendy/catkin_ws/build/realtime_tools; catkin build --get-env realtime_tools | catkin env -si  /usr/bin/cmake /home/jwhendy/catkin_ws/src/realtime_tools --no-warn-unused-cli -DCATKIN_DEVEL_PREFIX=/home/jwhendy/catkin_ws/devel/.private/realtime_tools -DCMAKE_INSTALL_PREFIX=/home/jwhendy/catkin_ws/install -DPYTHON_EXECUTABLE=/usr/bin/python2 -DPYTHON_INCLUDE_DIR=/usr/include/python2.7 -DPYTHON_LIBRARY=/usr/lib/libpython2.7.so; cd -
```

Also, I do have them installed:
```
$ pacman -Qi gmock
Name            : gmock
Version         : 1.8.1-3

$ pacman -Qi gtest
Name            : gtest
Version         : 1.8.1-3

$ ls /usr/include/gmock/
internal         gmock-cardinalities.h      gmock-generated-function-mockers.h  gmock-generated-nice-strict.h  gmock-matchers.h      gmock-more-matchers.h
[...]

PR #55, opened 2020-03-15
check for gmock before testing

acxz:
Due to some bug in either ArchLinux gtest/gmock packaging or catkin (still not sure where) `realtime_box_tests` is not able to build and causes a build error on ArchLinux systems.
In this repo, the flag `CATKIN_ENABLE_TESTING` is set to True even if `${GMOCK_MAIN_LIBRARIES}` does not exist. If `GMOCK_MAIN_LIBRARIES` is empty then the build will fail, instead of just completing without performing unit tests.

This PR creates a check for GMOCK_MAIN_LIBRARIES so that even if `CATKIN_ENABLE_TESTING` is true the tests will only go through if GMOCK_MAIN_LIBRARIES exist as well.

Relevant discussions: #44 here in this repo
https://github.com/ros/catkin/issues/1030   issue at catkin and
https://github.com/ros-melodic-arch/ros-melodic-realtime-tools/issues/1 at ArchLinux packaging.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -19,7 +19,7 @@ target_link_libraries(${PROJECT_NAME} PUBLIC ${catkin_LIBRARIES} ${CMAKE_THREAD_
 target_include_directories(${PROJECT_NAME} PUBLIC ${catkin_INCLUDE_DIRS})
 
 # Unit Tests
-if (CATKIN_ENABLE_TESTING)
+if (CATKIN_ENABLE_TESTING AND (EXISTS ${GMOCK_MAIN_LIBRARIES}))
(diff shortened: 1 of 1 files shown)


matthew-reynolds:
This seems like an upstream issue to me, not something we should be working around at the package level. Since `catkin` depends on `google-mock`, it seem to me that if `gmock` isn't found this is an issue with `catkin` or with Arch's packaging.

I don't fully understand all the conversation that went on in the linked threads, but my impression is that if people are seeing this issue with `realtime_tools`, they're bound to see it with any package using `catkin_add_gmock()`. Is that right?

acxz:
Yes you are correct, this issue should be tackled at the root, instead of here. I apologize for the PR.

PR #57, opened 2020-03-19
Realtime server goal thread handle safety + additional warning fixes (#2)

ddengster:
Some fixes to get ros2_controller's PR#26 working. I'm using createWallTimer to run the runNonRealtime() function at a rate 20Hz, and calling setFeedback sometimes makes these threads hit race conditions. 

Besides that the action server does not appreciate success/cancel/abort being sent multiple times; it issues a warning. This is also addressed by setting the booleans to false once sent.

Downstream PR: https://github.com/ros-controls/ros2_controllers/pull/26

### DIFF
--- a/include/realtime_tools/realtime_server_goal_handle.h
+++ b/include/realtime_tools/realtime_server_goal_handle.h
@@ -52,10 +52,10 @@ private:
 
   uint8_t state_;
 
-  std::atomic<bool> req_abort_;
-  std::atomic<bool> req_cancel_;
-  std::atomic<bool> req_succeed_;
-  std::atomic<bool> req_execute_;
+  bool req_abort_;
+  bool req_cancel_;
+  bool req_succeed_;
+  bool req_execute_;
 
   std::mutex mutex_;
   ResultSharedPtr req_result_;
--- a/include/realtime_tools/realtime_server_goal_handle.h
+++ b/include/realtime_tools/realtime_server_goal_handle.h
@@ -31,7 +31,6 @@
 #define REALTIME_TOOLS__REALTIME_SERVER_GOAL_HANDLE_H
 
 // Standard
-#include <atomic>
 #include <memory>
 
 #include <rclcpp/exceptions.hpp>
--- a/include/realtime_tools/realtime_server_goal_handle.h
+++ b/include/realtime_tools/realtime_server_goal_handle.h
@@ -31,6 +31,7 @@
 #define REALTIME_TOOLS__REALTIME_SERVER_GOAL_HANDLE_H
 
 // Standard
+#include <atomic>
 #include <memory>
 
 #include <rclcpp/exceptions.hpp>
@@ -51,10 +52,12 @@ private:
 
   uint8_t state_;
 
-  bool req_abort_;
-  bool req_cancel_;
-  bool req_succeed_;
-  bool req_execute_;
+  std::atomic<bool> req_abort_;
+  std::atomic<bool> req_cancel_;
+  std::atomic<bool> req_succeed_;
+  std::atomic<bool> req_execute_;
+
+  std::mutex mutex_;
   ResultSharedPtr req_result_;
   FeedbackSharedPtr req_feedback_;
   rclcpp::Logger logger_;
@@ -77,10 +80,10 @@ public:
       req_cancel_(false),
       req_succeed_(false),
       req_execute_(false),
+      logger_(logger),
       gh_(gh),
       preallocated_result_(preallocated_result),
-      preallocated_feedback_(preallocated_feedback),
-      logger_(logger)
+      preallocated_feedback_(preallocated_feedback)
   {
     if (!preallocated_result_)
       preallocated_result_.reset(new typename Action::Result);
@@ -92,6 +95,8 @@ public:
   {
     if (req_execute_ && !req_succeed_ && !req_abort_ && !req_cancel_)
     {
+      std::lock_guard<std::mutex> guard(mutex_);
+
       req_result_ = result;
       req_abort_ = true;
     }
@@ -101,6 +106,8 @@ public:
   {
     if (req_execute_ && !req_succeed_ && !req_abort_ && !req_cancel_)
     {
+      std::lock_guard<std::mutex> guard(mutex_);
+
       req_result_ = result;
       req_cancel_ = true;
     }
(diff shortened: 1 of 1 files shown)


include/realtime_tools/realtime_server_goal_handle.h
```diff
@@ -92,6 +95,8 @@ class RealtimeServerGoalHandle
   {
     if (req_execute_ && !req_succeed_ && !req_abort_ && !req_cancel_)
     {
+      std::lock_guard<std::mutex> guard(mutex_);
```
Karsten1987: If `runNonRealtime` is also locked with the same mutex, the `req_abort_` variable doesn't have to be an atomic. Same below.

include/realtime_tools/realtime_server_goal_handle.h
```diff
@@ -135,36 +146,41 @@ class RealtimeServerGoalHandle
...
-    if (valid())
+    if (!valid())
+      return;
+
+    std::lock_guard<std::mutex> guard(mutex_);
+
+    try
     {
-      try
+      if (req_execute_ && !gh_->is_executing() && gh_->is_active() && !gh_->is_canceling())
```
Karsten1987: I know it was in the original change already, but I have a hard time to believe that I can trigger `execute` on a goal which is in either of the other states ('executing, !active, canceling`). 

Also, would it make sense to check if the atomics are set here to guarantee a realtime behavior?

Karsten1987:
This generally looks good to me, however one question though I'd like to rule out before merging:

If the function `runNonRealtime` is supposed to run - as the name suggests - in non-realtime, does that mean all other functions are supposed to return in a realtime safe, deterministic time?

If so, I would request changes to all `lock_guards` in the other functions as they are potentially waiting for the `runNonRealTime` function to return, making them effectively non-realtime. In that case, all `lock_guards` should be replaced by some sort of `try_lock`.
```
auto sub_lock = std::unique_lock<std::mutex>(mutex_, std::try_to_lock);
if (sub_lock.owns_lock()) {
  //
}
``` 

If not, meaning the logs are really only there to avoid race conditions, I don't think we have to introduce `atomics` here as anyway all the calls are locked by the same mutex.

ddengster:
@Karsten1987 you're right this was supposed to be a PR that addressed several race conditions and bugs with the code and to get it working with the joint_trajectory action server PR. As to the realtime-safetyness of the code, it should be looked at in another PR.

include/realtime_tools/realtime_server_goal_handle.h
ddengster: there's a check for req_abort_ outside the lock_guard, so it should be atomic

include/realtime_tools/realtime_server_goal_handle.h
ddengster: see https://github.com/ros-controls/realtime_tools/pull/57#issuecomment-603634364

include/realtime_tools/realtime_server_goal_handle.h
```diff
@@ -135,36 +146,41 @@ class RealtimeServerGoalHandle
 
   void runNonRealtime()
   {
-    if (valid())
+    if (!valid())
+      return;
```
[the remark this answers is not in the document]
ddengster: yea it looked to me that the loop originally was meant to spin even if there's no goal handle. it might have to be reworked by decoupling the function from it's object.

include/realtime_tools/realtime_server_goal_handle.h
```diff
@@ -51,10 +52,12 @@ class RealtimeServerGoalHandle
...
   uint8_t state_;
 
-  bool req_abort_;
-  bool req_cancel_;
-  bool req_succeed_;
-  bool req_execute_;
+  std::atomic<bool> req_abort_;
+  std::atomic<bool> req_cancel_;
+  std::atomic<bool> req_succeed_;
+  std::atomic<bool> req_execute_;
```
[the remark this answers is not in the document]
ddengster: it's only set to false in the constructor.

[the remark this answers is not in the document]
Karsten1987: This doesn't allow to send a goal then, succeed it and then send it again. I would think this should be addressed.

include/realtime_tools/realtime_server_goal_handle.h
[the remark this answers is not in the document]
Karsten1987: If you could rule this out in this PR I would be happy. It seems to me as if that implementation is not clearly worked out in terms of usability. But that might just be my little knowledge about ros2 actions.

include/realtime_tools/realtime_server_goal_handle.h
Karsten1987: in the `if` condition? I mean that's linked with 3(!!) `&&` which allows for all sorts of context switches in between. So I don't see how an atomic could help here or the condition doesn't make sense.

include/realtime_tools/realtime_server_goal_handle.h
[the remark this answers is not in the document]
ddengster: > Or is this a one-time thing, where for each new call to the action server this goal will be newly created?

Yes. I've put in a new issue regarding this. https://github.com/ros-controls/ros2_controllers/issues/34

include/realtime_tools/realtime_server_goal_handle.h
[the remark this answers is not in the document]
ddengster: https://github.com/ros-controls/ros2_controllers/issues/34

include/realtime_tools/realtime_server_goal_handle.h
ddengster: fa6281a

ddengster:
@Karsten1987 any more blockers you're concerned about? i need this to fix the test upstream in ros2_controllers

include/realtime_tools/realtime_server_goal_handle.h
```diff
@@ -31,6 +31,7 @@
 #define REALTIME_TOOLS__REALTIME_SERVER_GOAL_HANDLE_H
 
 // Standard
+#include <atomic>
```
bmagyar, line 34: unused

ddengster: fixed 5fb2495

bmagyar: thank you

PR #59, opened 2020-04-16
Migrate to new industrial_ci

matthew-reynolds:
See also ros-controls/ros_control#440


matthew-reynolds:
Should be picked over to `noetic-devel` as well

PR #61, opened 2020-06-10
avoid deprecations

Karsten1987:
as per title


bmagyar:
The title isn't very descriptive sir. Could you please copy the deprecation warning for these next time?

PR #71, opened 2021-03-18
Fix deprecation warnings when constructing rclcpp::Duration

jacobperron:
Since https://github.com/ros2/rclcpp/pull/1432 (upcoming in Galactic), we should not initialize with a single integer
as the units are ambiguous.

---

I wasn't sure if I should target `foxy-devel` or the `ros2_devel` branch. I picked `foxy-devel` since it seems the most up-to-date. This fix is compatible with Foxy.


bmagyar:
Nice catch, thanks!

PR #76, opened 2021-06-03
Missing main function for running all tests

vignif:
Missing main snippet


bmagyar:
I think this would cause issues as we are already linking the gmock main to do exactly that.

In the cmake you'll find `${GMOCK_MAIN_LIBRARIES}`

vignif:
Oo thanks for the feedback, I didn't notice that!

PR #79, opened 2021-09-21
Accommodate the RCLCPP_WARN macro by using a C string

mjeronimo:
Fix a compile-time issue with the latest ROS 2 code.

Signed-off-by: Michael Jeronimo <michael.jeronimo@openrobotics.org>


bmagyar:
This should be fine on Foxy and Galactic as well, right?

mjeronimo:
@bmagyar Yes, the char * version will work on either. The std::string version doesn't work for Galactic (and Rolling).

bmagyar:
Um... How come you found out about this issue? It's apparently already fixed on Foxy

mjeronimo:
I was using the ros2_devel branch with Rolling.

PR #83, opened 2022-08-02
Add a library to detect a realtime kernel and set thread priority

AndyZe:
As suggested in this ros2_control PR:  https://github.com/ros-controls/ros2_control/pull/792

Add a library to detect a realtime kernel and set thread priority. Having it here should make it easy to link against from other packages and generally seems like a better place to have the library than inside `controller_manager`.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -18,10 +18,7 @@ target_include_directories(${PROJECT_NAME} PUBLIC include)
 
 # A library to detect a realtime kernel and set thread priority, if one is found
 add_library(thread_priority STATIC src/thread_priority.cpp)
-target_include_directories(thread_priority PUBLIC
-  "$<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}/include>"
-  "$<INSTALL_INTERFACE:include>"
-)
+target_include_directories(thread_priority PUBLIC include)
 ament_target_dependencies(thread_priority ${THIS_PACKAGE_INCLUDE_DEPENDS})
 
 # Unit Tests
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -16,6 +16,14 @@ ament_target_dependencies(${PROJECT_NAME} "rclcpp" "rclcpp_action")
 target_link_libraries(${PROJECT_NAME} ${CMAKE_THREAD_LIBS_INIT})
 target_include_directories(${PROJECT_NAME} PUBLIC include)
 
+# A library to detect a realtime kernel and set thread priority, if one is found
+add_library(thread_priority STATIC src/thread_priority.cpp)
+target_include_directories(thread_priority PUBLIC
+  "$<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}/include>"
+  "$<INSTALL_INTERFACE:include>"
+)
+ament_target_dependencies(thread_priority ${THIS_PACKAGE_INCLUDE_DEPENDS})
+
 # Unit Tests
 if(BUILD_TESTING)
   find_package(ament_cmake_gmock REQUIRED)
@@ -44,13 +52,13 @@ install(
   DIRECTORY include/
   DESTINATION include)
 
-install(TARGETS ${PROJECT_NAME}
+install(TARGETS ${PROJECT_NAME} thread_priority
   ARCHIVE DESTINATION lib
   LIBRARY DESTINATION lib
   RUNTIME DESTINATION bin)
 
 ament_export_include_directories(include)
-ament_export_libraries(${PROJECT_NAME})
+ament_export_libraries(${PROJECT_NAME} thread_priority)
 
 ament_export_dependencies(rclcpp)
 ament_export_dependencies(rclcpp_action)
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -13,13 +13,13 @@ find_package(Threads REQUIRED)
 
 add_library(${PROJECT_NAME} src/realtime_clock.cpp)
 ament_target_dependencies(${PROJECT_NAME} "rclcpp" "rclcpp_action")
-target_link_libraries(${PROJECT_NAME} ${CMAKE_THREAD_LIBS_INIT})
+target_link_libraries(${PROJECT_NAME} ${Update CMakeLists.txt})
 target_include_directories(${PROJECT_NAME} PUBLIC include)
 
 # A library to detect a realtime kernel and set thread priority, if one is found
 add_library(thread_priority STATIC src/thread_priority.cpp)
 target_include_directories(thread_priority PUBLIC include)
-ament_target_dependencies(thread_priority ${THIS_PACKAGE_INCLUDE_DEPENDS})
+ament_target_dependencies(thread_priority "rclcpp" "rclcpp_action")
 
 # Unit Tests
 if(BUILD_TESTING)
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -13,7 +13,7 @@ find_package(Threads REQUIRED)
 
 add_library(${PROJECT_NAME} src/realtime_clock.cpp)
 ament_target_dependencies(${PROJECT_NAME} "rclcpp" "rclcpp_action")
-target_link_libraries(${PROJECT_NAME} ${Update CMakeLists.txt})
+target_link_libraries(${PROJECT_NAME} ${CMAKE_THREAD_LIBS_INIT})
 target_include_directories(${PROJECT_NAME} PUBLIC include)
 
 # A library to detect a realtime kernel and set thread priority, if one is found
--- a/include/realtime_tools/thread_priority.hpp
+++ b/include/realtime_tools/thread_priority.hpp
@@ -1,0 +1,50 @@
+/*
+ * Copyright (c) 2022 PickNik Inc.
+ * All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are met:
+ *
+ *     * Redistributions of source code must retain the above copyright
+ *       notice, this list of conditions and the following disclaimer.
+ *     * Redistributions in binary form must reproduce the above copyright
+ *       notice, this list of conditions and the following disclaimer in the
+ *       documentation and/or other materials provided with the distribution.
+ *     * Neither the name of the Willow Garage, Inc. nor the names of its
+ *       contributors may be used to endorse or promote products derived from
+ *       this software without specific prior written permission.
+ *
+ * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS"
+ * AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE
+ * IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE
+ * ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT OWNER OR CONTRIBUTORS BE
+ * LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR
+ * CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF
+ * SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS
+ * INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN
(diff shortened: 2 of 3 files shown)


CMakeLists.txt
```diff
@@ -11,11 +11,18 @@ find_package(rclcpp REQUIRED)
 find_package(rclcpp_action REQUIRED)
 find_package(Threads REQUIRED)
 
+include_directories(include)
```
tylerjw, line 14: Using the global function `include_directories` is considered bad practice in modern cmake. Please use `target_include_directories` and export the library as part of the install set.

CMakeLists.txt
```diff
@@ -11,11 +11,18 @@ find_package(rclcpp REQUIRED)
...
+
 add_library(${PROJECT_NAME} src/realtime_clock.cpp)
 ament_target_dependencies(${PROJECT_NAME} "rclcpp" "rclcpp_action")
 target_link_libraries(${PROJECT_NAME} ${CMAKE_THREAD_LIBS_INIT})
 target_include_directories(${PROJECT_NAME} PUBLIC include)
 
+# A library to detect a realtime kernel and set thread priority, if one is found
+add_library(thread_priority SHARED src/thread_priority.cpp)
```
tylerjw, line 22: Why did you make this a shared library?

include/realtime_tools/thread_priority.hpp
```diff
@@ -0,0 +1,50 @@
...
+ * CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF
+ * SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS
+ * INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN
+ * CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE)
+ * ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+ * POSSIBILITY OF SUCH DAMAGE.
+ */
+
+#ifndef CONTROLLER_MANAGER__REALTIME_HPP_
+#define CONTROLLER_MANAGER__REALTIME_HPP_
```
tylerjw, line 31: These header guards are now wrong.

CMakeLists.txt
AndyZe: i don't have a strong opinion. Execution time shouldn't really matter, so I figured SHARED would be a bit more memory-efficient

CMakeLists.txt
AndyZe: done

include/realtime_tools/thread_priority.hpp
AndyZe: done

CMakeLists.txt
tylerjw: What part of using a shared library is more memory efficient? Are you trying to reduce the memory used by the linker?

CMakeLists.txt
```diff
@@ -16,6 +16,11 @@ ament_target_dependencies(${PROJECT_NAME} "rclcpp" "rclcpp_action")
...
+target_include_directories(thread_priority PRIVATE include)
```
tylerjw, line 21: This is probably wrong. Have you tried using this somewhere?

AndyZe: Yep, this PR from ros2_control:  https://github.com/ros-controls/ros2_control/pull/794

CMakeLists.txt
AndyZe: **As well as leading to smaller applications**, shared libraries offer the user the ability to use their own, perhaps better version of the libraries rather than relying on one that's part of the application

https://stackoverflow.com/a/2649379/3499467

Again, I don't really care so much. If you have a strong reason to make it a static library, that's fine with me.

tylerjw: I think static libraries should be the default unless you are creating a plugin or something like openssl that has the requirements of being updated for security reasons.

CMakeLists.txt
tylerjw: I don't understand why this would work because PRIVATE means that these includes are not part of your exported interface, when clearly they are.

tylerjw: I think normally this should be:
```cmake
target_include_directories(thread_priority
  $<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}/include>
  $<INSTALL_INTERFACE:${CMAKE_INSTALL_INCLUDEDIR}>)
```

tylerjw: PRIVATE here makes sense in the ros2_control plugins that are not libraries because they don't export headers as they are just plugins.

AndyZe: Your suggestion -->

CMake Error at CMakeLists.txt:21 (target_include_directories):
  target_include_directories called with invalid arguments

I'm looking into it

AndyZe: This works:

```
target_include_directories(thread_priority PUBLIC
  "$<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}/include>"
  "$<INSTALL_INTERFACE:include>"
)
```

destogl: I think this should be as it was, but with "PUBLIC". I was also using this complex definitions few times, but never found out where they actually have to be used. I mean why make CMakeLists.txt less readable if there is no advantage of it.

ChrisThrasher: That generator expression ensures that when building the library you get the proper public include directory but it also ensures that when you install the library, the installed library doesn't include `/home/my_user/Repos/realtime_tools/include` since such a user-specific include directory makes no sense for shipping libraries and is going to break even on your own machine should you ever delete the source code repo from which you built the library.

Yes, the more complicated form looks very tedious but the simple `target_include_directories(libName PUBLIC include)` form is insufficient for library targets that need to get installed.

CMakeLists.txt
```diff
@@ -16,6 +16,14 @@ ament_target_dependencies(${PROJECT_NAME} "rclcpp" "rclcpp_action")
 target_link_libraries(${PROJECT_NAME} ${CMAKE_THREAD_LIBS_INIT})
 target_include_directories(${PROJECT_NAME} PUBLIC include)
 
+# A library to detect a realtime kernel and set thread priority, if one is found
+add_library(thread_priority STATIC src/thread_priority.cpp)
+target_include_directories(thread_priority PUBLIC
+  "$<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}/include>"
+  "$<INSTALL_INTERFACE:include>"
+)
```
destogl, line 24: why not simply:

```suggestion
target_include_directories(thread_priority PUBLIC include)
```

or will this not work for some reason? I am using this all the time and didn't have any issue until now.

include/realtime_tools/thread_priority.hpp
```diff
@@ -0,0 +1,47 @@
...
+ * LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR
+ * CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF
+ * SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS
+ * INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN
+ * CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE)
+ * ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+ * POSSIBILITY OF SUCH DAMAGE.
+ */
+
+#pragma once
```
destogl, line 30:
```suggestion
#ifndef REALTIME_TOOLS__JOINT_THREAD_PRIORITY_HPP_
#define REALTIME_TOOLS__JOINT_THREAD_PRIORITY_HPP_
```

include/realtime_tools/thread_priority.hpp
```diff
@@ -0,0 +1,47 @@
...
+bool has_realtime_kernel();
+
+/**
+ * Configure SCHED_FIFO thread priority for the thread that calls this function
+ * \param[in] priority the priority of this thread from 0-99
+ * \returns true if configuring scheduler succeeded
+ */
+bool configure_sched_fifo(int priority);
+
+}  // namespace realtime_tools
```
destogl, line 47:
```suggestion
}  // namespace realtime_tools

#endif  // REALTIME_TOOLS__JOINT_THREAD_PRIORITY_HPP_

```

src/thread_priority.cpp
```diff
@@ -0,0 +1,58 @@
...
+ * ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+ * POSSIBILITY OF SUCH DAMAGE.
+ */
+
+#include "realtime_tools/thread_priority.hpp"
+
+#include <sched.h>
+
+#include <cstring>
+#include <fstream>
```
destogl, line 35:
```suggestion
#include <cstring>
#include <fstream>
#include <sched.h>
```

destogl (changes requested):
Thanks, just a few minor comments. Guard it the main difference.

CMakeLists.txt
destogl: I have different experience. We are using simpler form in whole ros2_control, which are basically many libraries and people don't have any problems executing them. I understand, but can not confirm you explanation from my experience. Especially, look at few lines above that we use simple from which is used in *every* controller in `ros2_controllers` repository and we have never had any issues you described.

Is this because we are using ROS build farm?

include/realtime_tools/thread_priority.hpp
AndyZe: done

CMakeLists.txt
AndyZe: If I interpret what @ChrisThrasher wrote correctly, you wouldn't see any broken functionality from `target_include_directories(libName PUBLIC include)`, but you might get some extra junk installed. Only if you have already built from source on the same machine, though.

I trust him a lot, he's the PickNik expert on CMake so I won't change this.

CMakeLists.txt
AndyZe: Not planning to change this. See the discussion above.

bmagyar (approved):
I'm happy for this to go ahead when you guys are happy

src/thread_priority.cpp
```diff
@@ -0,0 +1,57 @@
...
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are met:
+ *
+ *     * Redistributions of source code must retain the above copyright
+ *       notice, this list of conditions and the following disclaimer.
+ *     * Redistributions in binary form must reproduce the above copyright
+ *       notice, this list of conditions and the following disclaimer in the
+ *       documentation and/or other materials provided with the distribution.
+ *     * Neither the name of the Willow Garage, Inc. nor the names of its
```
bmagyar, line 13: Willow?

include/realtime_tools/thread_priority.hpp
```diff
@@ -0,0 +1,50 @@
...
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are met:
+ *
+ *     * Redistributions of source code must retain the above copyright
+ *       notice, this list of conditions and the following disclaimer.
+ *     * Redistributions in binary form must reproduce the above copyright
+ *       notice, this list of conditions and the following disclaimer in the
+ *       documentation and/or other materials provided with the distribution.
+ *     * Neither the name of the Willow Garage, Inc. nor the names of its
```
bmagyar, line 13: Willow

CMakeLists.txt
```diff
@@ -13,9 +13,14 @@ find_package(Threads REQUIRED)
 
 add_library(${PROJECT_NAME} src/realtime_clock.cpp)
 ament_target_dependencies(${PROJECT_NAME} "rclcpp" "rclcpp_action")
-target_link_libraries(${PROJECT_NAME} ${CMAKE_THREAD_LIBS_INIT})
+target_link_libraries(${PROJECT_NAME} ${Update CMakeLists.txt})
```
destogl, line 16:
```suggestion
target_link_libraries(${PROJECT_NAME} ${CMAKE_THREAD_LIBS_INIT})
```

include/realtime_tools/thread_priority.hpp
bmagyar:
```suggestion
 *     * Neither the name of the PickNik Inc. nor the names of its
```

src/thread_priority.cpp
bmagyar:
```suggestion
 *     * Neither the name of the PickNik Inc. nor the names of its
```

PR #128, opened 2023-11-07
initialize the global context to avoid runtime_error upon destruction

saikishor:
This fixes the issue pointed by @christophfroehlich about the failing tests in the rolling that at the end of the tests upon destruction, the context becomes invalid

```
[ RUN ] RealtimeClock.get_system_time
 terminate called after throwing an instance of 'std::runtime_error'
 what(): context cannot be slept with because it's invalid
 ```


christophfroehlich (approved):
A simple fix, thanks for digging into this!

bmagyar (approved):
Thank you!

PR #129, opened 2023-11-07
[CI] Add jobs for iron

christophfroehlich:
and also update the table in the readme.

fails until #128 is merged


bmagyar:
Let's see now how Everything Works TM

christophfroehlich:
Seems to run fine. Also updated the readme.

PR #133, opened 2023-11-14
Fix ignore and patch for codecov

christophfroehlich:
The default branch on codecov is "indigo-devel", I cannot change it: 

> We were unable to update the default branch for this repo

I also fixed the glob for the ignore of the test folder.


bmagyar (approved):
Thank you, I've updated the default branch to be `master`

Issue #138, opened 2024-01-12
Comment in realtime_box might be wrong and misleading


firesurfer:
Hi, as I am currently writing some controllers I started to take a look at the realtime_tools.

The first comment in the realtime_box states:

>   Strongly suggested that you use an std::shared_ptr in this box to guarantee realtime safety.

https://github.com/ros-controls/realtime_tools/blob/441b544ea32afea6658341a50dbadc149a41368a/include/realtime_tools/realtime_box.h#L41

In my opinion one should not use pointers at all with the current implementation of the RealtimeBox as we do:

1. Obtain a lock
2. Copy the stored value to the passed reference
3. Unlock the lock 
4. And now the user will start operating on the pointer

The lock is already released when operations on the actual data are done. 

Perhaps my understanding of the specific case in incomplete (please correct me in that case). 
But I would suggest to actually change the comment in: `Using pointers is not safe with this implementation`

bmagyar:
Agreed. That's just wrong

Issue #81, opened 2022-02-25
realtime_tools API restructure


bmagyar:
The ROS2 realtime_tools API is a little clunky, requiring the user to still define the publisher on their side instead of incorporating it into the realtime publisher. 

This makes for a fairly defunct API where we don't gain much code reduction but there's still an increase in complexity and dependencies.

Rework the realtime_tools RealtimePublisher API such that it is as lean as possible to use.

YoavFekete:
i am in for the task

christophfroehlich:
@saikishor what do you think, could we move the creation of the publisher into the rt_publisher implementation?

PR #139, opened 2024-01-12
Added a new implementation of the RealtimeBox with added best effort behaviour

firesurfer:
Hi, I realized the current implementation of the RealtimeBox does not have a "try write/try read" access method. This im important from my point of view. Therefore I implemented a `RealtimeBoxBestEffort` which per default trys to call `try_lock` for read and write operations but also implements the previous behaviour which waits until it can aquire a lock. 

As already stated in #138 this implementation is not safe to be used with pointers from my point of view. (In the tests I added an example which shows how to use a pointer type with the box but well..for pointers not using this kind of box might be better or we could specialize the template to provide a different behaviour)

Furthermore I took the chance to provide some more modern c++ "magic". For example it is possible to have "emplace behaviour" - basically construct in place or  if the underlying type allows it to directly pass an initilizer list.

_Note:_ This implementation requires at least c++17 which is the current default target for ROS2 if I am informed correctly.

Perhaps we could even figure out a way how to replace the current RealTimeBox with an updated interface (#81 says that there are plans to update the API in general)


bmagyar:
@christophfroehlich here we have a new contender for C++17 support. Do you know how Debian compilation would be affected?

christophfroehlich:
> @christophfroehlich here we have a new contender for C++17 support. Do you know how Debian compilation would be affected?

it compiles and the test runs successfully on debian11

firesurfer:
@christophfroehlich  I reworked the PR and you might want to test if it still compiles on older platforms.

1. I restructured the interface. It should now be compatible with the existing RealtimeBox.
2. I added a way to safely use the Box with pointer types. (If you come up with a better way let me know). See the last two tests for an example.
3. For pointer types the normal get/set methods are disabled via std::enable_if as they are not safe to use.

As further steps I would suggest:
1. I would replace the existing RT Box with the new implementation and see if the existing tests work fine with it
2. Add some documentation about the usage (Is there any prefered place in the repository to store that?)

christophfroehlich:
I can't give proper feedback to your implementation, but it still compiles on debian11. And I made a mistake pushing something of your PR to the master (that's why the conflicts now appeared)

btw: I propose that we fix the RHEL image and add maybe another CI for the supported versions of debian to our CIs

firesurfer:
Then given that it still compiles for you on an older system perhaps @bmagyar  might want to review it ? 

I am btw not sure why the formatting step fails. I am pretty sure I ran clang-format on all files I touched.

christophfroehlich:
> I am btw not sure why the formatting step fails. I am pretty sure I ran clang-format on all files I touched.

The copyright notice is missing in the files, and maybe you have different clang settings. Please install pre-commit in this repo and run it (`pre-commit run --all`)

> EDIT: Perhaps you might want to note at a prominent place that you have additional requirements for this package when it comes to build platforms. As far as I know the base line for rolling/iron is Ubuntu 22.04 and C++17 (correct me if I am wrong)

The mentioned combination was never a problem so far, but gcc on debian11 (which is [Tier 3](https://www.ros.org/reps/rep-2000.html#humble-hawksbill-may-2022-may-2027) for humble) has not implemented all of the c++17 standard. This is why `std::from_chars` is not supported, what caused problems with ros2_control.

firesurfer:
@saikishor  Do you mean the style of the methods themselves or of the comments? 
@christophfroehlich Would it be possible for you to resolve the merge conflicts? I am not sure what exactly the best way would be to resolve them. 

Is there anything else required to get this merged? 

I would additionally add another PR to replace the existing RealtimeBox implementation with this one. I already checked locally on my system if all RTBox Tests also run fine with this implementation.

christophfroehlich:
@firesurfer please check if the merge was successful, and why the format job fails again.

christophfroehlich:
> @christophfroehlich Should be fixed now. The formatter just added a whitespace at the end of some comments...
> 
> I think cppcheck can't cope wit the std::enable_if_t at two points which is why the ament_cppcheck fails
> 
> ```
>  /usr/bin/bash -c source /opt/ros/rolling/setup.sh && ament_cppcheck $(colcon list --packages-select realtime_tools -p) --language=c++
> Error: [include/realtime_tools/realtime_box_best_effort.h:141]: (error: missingReturn) Found a exit path from function with non-void return type that has missing return statement
> Error: [include/realtime_tools/realtime_box_best_effort.h:170]: (error: missingReturn) Found a exit path from function with non-void return type that has missing return statement
> ```

This seems to be fixed and should be part of [cppcheck 2.10](https://trac.cppcheck.net/ticket/11171), but this is not released for [Ubuntu jammy](https://packages.ubuntu.com/jammy/cppcheck). Nothing we can do here?

format job is still complaining:
`include/realtime_tools/realtime_box_best_effort.h:204: carefull ==> careful, carefully`

firesurfer:
I fixed the format job now (hopefully). Sorry that this took so long. For some reason I get way more errors when I run `pre-commit run --all` locally. Especially `ament_cpplint` complains about quite a lot of formatting but I guess this is disabled in the ci job?

firesurfer:
@christophfroehlich Apparently now the ament_cpplint fails. 

I am a bit puzzled because at least half of the issues there should be fixed by clang-format as they are purely formatting related. What is the best way to fix them without going manually through each of them.

I already tried ament_uncrustify but then the clang-format check from pre-commit fails again.

include/realtime_tools/realtime_box_best_effort.h
```diff
@@ -0,0 +1,216 @@
...
+
+// Author: Lennart Nachtigall
+
+#pragma once
+
+#include <initializer_list>
+#include <mutex>
+#include <optional>
+#include <rcpputils/pointer_traits.hpp>
+#include <utility>
```
christophfroehlich, line 37:
```suggestion
#include <initializer_list>
#include <mutex>
#include <optional>
#include <utility>

#include <rcpputils/pointer_traits.hpp>
```

I think it should look like this

firesurfer:
@christophfroehlich  All checks seem to pass now. I simply suppressed the two cppcheck warnings regarding the wrong return type.

firesurfer:
@christophfroehlich So all checks have passed. What is needed to get this merged?

christophfroehlich:
A review from @bmagyar or @destogl ;)

firesurfer:
@bmagyar  or @destogl would you mind to review the changes?

include/realtime_tools/realtime_box_best_effort.h
```diff
@@ -0,0 +1,220 @@
...
+// CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF
+// SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS
+// INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN
+// CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE)
+// ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+// POSSIBILITY OF SUCH DAMAGE.
+
+// Author: Lennart Nachtigall
+
+#pragma once
```
destogl, line 31: Please use `ifndef` header protections like here: https://github.com/ros-controls/realtime_tools/blob/master/include/realtime_tools/realtime_publisher.h#L37-L38

include/realtime_tools/realtime_box_best_effort.h
```diff
@@ -0,0 +1,220 @@
...
+    std::enable_if_t<std::is_constructible_v<U, std::initializer_list>>)
+  : value_(init)
+  {
+  }
+
+  /**
+     * @brief set a new content with best effort
+     * @return false if mutex could not be locked
+     * @note disabled for pointer types
+    */
```
destogl, line 81: Alignment here looks strange a bit.

include/realtime_tools/realtime_box_best_effort.h
```diff
@@ -0,0 +1,220 @@
...
+    if (!guard.try_lock()) {
+      return false;
+    }
+
+    func(value_);
+    return true;
+  }
+
+  /**
+     * @brief set the content and wait until the mutex could be locked (RealtimeBox behaviour)
```
destogl, line 137:
```suggestion
     * @brief set the content and wait until the mutex could be locked (RealtimeBox behavior)
```

include/realtime_tools/realtime_box_best_effort.h
```diff
@@ -0,0 +1,220 @@
...
+  template <typename U = T>
+  typename std::enable_if_t<!is_ptr_or_smart_ptr<U>, void> set(const T & value)
+  {
+    std::lock_guard<std::mutex> guard(lock_);
+    // cppcheck-suppress missingReturn
+    value_ = value;
+  }
+  /**
+   * @brief access the content (rw) and wait until the mutex could locked
+  */
```
destogl, line 149: I think that this, as here with my proposal is, is a correct alignment.
```suggestion
   */
```

include/realtime_tools/realtime_box_best_effort.h
```diff
@@ -0,0 +1,220 @@
...
+  void get(const std::function<void(const T &)> & func)
+  {
+    std::lock_guard<std::mutex> guard(lock_);
+    func(value_);
+  }
+
+  // Make the usage easier by providing a custom assignment operator
+  // and a custom conversion operator
+  // Only to be used from non-RT!
```
destogl, line 190: why not define this as docstring?

destogl (approved):
This looks good! Just a few comments about formatting and style

include/realtime_tools/realtime_box_best_effort.h
firesurfer: Even though I really try to avoid include guards I will change this in order to be compliant with the rest of the code

include/realtime_tools/realtime_box_best_effort.h
firesurfer: Because this applies to all the following operators and not just the first one. I will separate it into a doc string per operator.

firesurfer:
@destogl  I tried to resolve your comments wherever possible. Most of the formatting comments are not possible to resolve because `clang-format` decides to format the comments that way.

firesurfer:
So anyone interested in clicking on merge ;) As far is can tell there should be no remaining issue?

firesurfer:
@destogl @christophfroehlich Is there any remaining issue left in order to get this merged?

christophfroehlich (approved):

include/realtime_tools/realtime_box_best_effort.h
```diff
@@ -0,0 +1,232 @@
...
+#include <optional>
+#include <utility>
+
+#include <rcpputils/pointer_traits.hpp>
+
+namespace realtime_tools
+{
+
+template <typename T>
+constexpr auto is_ptr_or_smart_ptr = rcpputils::is_pointer<T>::value;
```
saikishor, line 45: `is_pointer` trait already checks if it is a poiner or smart pointer. Do we need to redefine it?

include/realtime_tools/realtime_box_best_effort.h
```diff
@@ -0,0 +1,232 @@
...
+
+  /**
+   * @brief set a new content with best effort
+   * @return false if mutex could not be locked
+   * @note disabled for pointer types
+   */
+  template <typename U = T>
+  typename std::enable_if_t<!is_ptr_or_smart_ptr<U>, bool> trySet(const T & value)
+  {
+    std::unique_lock<std::mutex> guard(lock_, std::defer_lock);
```
saikishor, line 86:
```suggestion
    std::unique_lock<mutex_t> guard(lock_, std::defer_lock);
```

include/realtime_tools/realtime_box_best_effort.h
```diff
@@ -0,0 +1,232 @@
...
+    return true;
+  }
+  /**
+   * @brief access the content readable with best effort
+   * @return false if the mutex could not be locked
+   * @note only safe way to access pointer type content (rw)
+   */
+  bool trySet(const std::function<void(T &)> & func)
+  {
+    std::unique_lock<std::mutex> guard(lock_, std::defer_lock);
```
saikishor, line 100:
```suggestion
    std::unique_lock<mutex_t> guard(lock_, std::defer_lock);
```

include/realtime_tools/realtime_box_best_effort.h
```diff
@@ -0,0 +1,232 @@
...
+    std::lock_guard<std::mutex> guard(lock_);
+    func(value_);
+  }
+
+  /**
+   * @brief provide a custom assignment operator for easier usage
+   * @note only to be used from non-RT!
+   */
+  template <typename U = T>
+  typename std::enable_if_t<!is_ptr_or_smart_ptr<U>, void> operator=(const T & value)
```
saikishor, line 194:
```suggestion
  typename std::enable_if_t<!is_ptr_or_smart_ptr<T>, void> operator=(const T & value)
```

include/realtime_tools/realtime_box_best_effort.h
```diff
@@ -0,0 +1,232 @@
...
+  {
+  }
+
+  /**
+   * @brief set a new content with best effort
+   * @return false if mutex could not be locked
+   * @note disabled for pointer types
+   */
+  template <typename U = T>
+  typename std::enable_if_t<!is_ptr_or_smart_ptr<U>, bool> trySet(const T & value)
```
saikishor, line 84:
```suggestion
  typename std::enable_if_t<!is_ptr_or_smart_ptr<T>, bool> trySet(const T & value)
```

include/realtime_tools/realtime_box_best_effort.h
```diff
@@ -0,0 +1,232 @@
...
+
+    func(value_);
+    return true;
+  }
+  /**
+   * @brief get the content with best effort
+   * @return std::nullopt if content could not be access, otherwise the content is returned
+   */
+  template <typename U = T>
+  [[nodiscard]] typename std::enable_if_t<!is_ptr_or_smart_ptr<U>, std::optional<U>> tryGet() const
```
saikishor, line 113:
```suggestion
  [[nodiscard]] typename std::enable_if_t<!is_ptr_or_smart_ptr<T>, std::optional<T>> tryGet() const
```

@firesurfer Do we need this typenames here?

include/realtime_tools/realtime_box_best_effort.h
```diff
@@ -0,0 +1,232 @@
...
+  {
+    std::unique_lock<std::mutex> guard(lock_, std::defer_lock);
```
saikishor, line 115:
```suggestion
    std::unique_lock<mutex_t> guard(lock_, std::defer_lock);
```

include/realtime_tools/realtime_box_best_effort.h
```diff
@@ -0,0 +1,232 @@
...
+    return value_;
+  }
+  /**
+   * @brief access the content (r) with best effort
+   * @return false if the mutex could not be locked
+   * @note only safe way to access pointer type content (r)
+   */
+  bool tryGet(const std::function<void(const T &)> & func)
+  {
+    std::unique_lock<std::mutex> guard(lock_, std::defer_lock);
```
saikishor, line 128:
```suggestion
    std::unique_lock<mutex_t> guard(lock_, std::defer_lock);
```

include/realtime_tools/realtime_box_best_effort.h
```diff
@@ -0,0 +1,232 @@
...
+    func(value_);
+    return true;
+  }
+
+  /**
+   * @brief set the content and wait until the mutex could be locked (RealtimeBox behavior)
+   * @return true
+   */
+  template <typename U = T>
+  typename std::enable_if_t<!is_ptr_or_smart_ptr<U>, void> set(const T & value)
```
saikishor, line 142:
```suggestion
  typename std::enable_if_t<!is_ptr_or_smart_ptr<T>, void> set(const T & value)
```

include/realtime_tools/realtime_box_best_effort.h
```diff
@@ -0,0 +1,232 @@
...
+  {
+    std::lock_guard<std::mutex> guard(lock_);
```
saikishor, line 144:
```suggestion
    std::lock_guard<mutex_t> guard(lock_);
```

include/realtime_tools/realtime_box_best_effort.h
```diff
@@ -0,0 +1,232 @@
...
+    // cppcheck-suppress missingReturn
+    value_ = value;
+  }
+  /**
+   * @brief access the content (rw) and wait until the mutex could locked
+   */
+  void set(const std::function<void(T &)> & func)
+  {
+    std::lock_guard<std::mutex> guard(lock_);
```
saikishor, line 153:
```suggestion
    std::lock_guard<mutex_t> guard(lock_);
```

include/realtime_tools/realtime_box_best_effort.h
```diff
@@ -0,0 +1,232 @@
...
+  }
+
+  /**
+   * @brief get the content and wait until the mutex could be locked (RealtimeBox behaviour)
+   * @return copy of the value
+   */
+  template <typename U = T>
+  [[nodiscard]] typename std::enable_if_t<!is_ptr_or_smart_ptr<U>, U> get() const
+  {
+    std::lock_guard<std::mutex> guard(lock_);
```
saikishor, line 164:
```suggestion
    std::lock_guard<mutex_t> guard(lock_);
```

include/realtime_tools/realtime_box_best_effort.h
```diff
@@ -0,0 +1,232 @@
...
+    func(value_);
+  }
+
+  /**
+   * @brief get the content and wait until the mutex could be locked (RealtimeBox behaviour)
+   * @return copy of the value
+   */
+  template <typename U = T>
+  [[nodiscard]] typename std::enable_if_t<!is_ptr_or_smart_ptr<U>, U> get() const
```
saikishor, line 162:
```suggestion
  [[nodiscard]] typename std::enable_if_t<!is_ptr_or_smart_ptr<T>, T> get() const
```

include/realtime_tools/realtime_box_best_effort.h
```diff
@@ -0,0 +1,232 @@
...
+  {
+    std::lock_guard<std::mutex> guard(lock_);
+    return value_;
+  }
+  /**
+   * @brief get the content and wait until the mutex could be locked
+   * @note same signature as in the existing RealtimeBox<T>
+   */
+  template <typename U = T>
+  typename std::enable_if_t<!is_ptr_or_smart_ptr<U>, void> get(T & in) const
```
saikishor, line 172:
```suggestion
  typename std::enable_if_t<!is_ptr_or_smart_ptr<T>, void> get(T & in) const
```

include/realtime_tools/realtime_box_best_effort.h
```diff
@@ -0,0 +1,232 @@
...
+  {
+    std::lock_guard<std::mutex> guard(lock_);
```
saikishor, line 174:
```suggestion
    std::lock_guard<mutex_t> guard(lock_);
```

include/realtime_tools/realtime_box_best_effort.h
```diff
@@ -0,0 +1,232 @@
...
+    in = value_;
+  }
+  /**
+   * @brief access the content (r) and wait until the mutex could be locked
+   * @note only safe way to access pointer type content (r)
+   * @note same signature as in the existing RealtimeBox<T>
+   */
+  void get(const std::function<void(const T &)> & func)
+  {
+    std::lock_guard<std::mutex> guard(lock_);
```
saikishor, line 185:
```suggestion
    std::lock_guard<mutex_t> guard(lock_);
```

test/realtime_box_best_effort_tests.cpp
```diff
@@ -0,0 +1,175 @@
...
+
+  RealtimeBoxBestEffort box(ptr);
+  // This does not and should not compile!
+  // auto value = box.get();
+
+  // Instead access it via a passed function.
+  // This assures that we access the data within the scope of the lock
+  box.get([](const auto & i) { EXPECT_EQ(*i, 100); });
+
+  box.set([](auto & i) { *i = 200; });
```
saikishor, line 153: @firesurfer Can you also add tests for the `trySet` method. I belive it shouldn't be a lot different. It's just to improve coverage

saikishor (changes requested):
Few minor suggestions.

Thank you! :)

include/realtime_tools/realtime_box_best_effort.h
firesurfer: I guess something like  `is_pointer_v = rcpputils::is_pointer<T>::value` would make more sense. 
The reason I added it in the first place was because I wasn't aware of the rcpputils::is_pointer in the beginning.

include/realtime_tools/realtime_box_best_effort.h
firesurfer: You are absolutely right about this one. Thats something I missed.

include/realtime_tools/realtime_box_best_effort.h
firesurfer: In order to use SFINAE to enable/disable member functions we need to have the additional function template

include/realtime_tools/realtime_box_best_effort.h
firesurfer: Just looked it up. It makes sense to call it like that as there is: https://en.cppreference.com/w/cpp/types/is_pointer
Which only checks if a type is a raw pointer

firesurfer:
@saikishor Thanks for your comments. I implemented most of them as appropriate. (E.g. we can not remove the additional function template parameter). The only open point is if we shall rename / remove `is_ptr_or_smart_ptr`. Just see the conversation above.

bmagyar:
Only waiting on CI and a thumbs up from @saikishor

include/realtime_tools/realtime_box_best_effort.h
saikishor: Hello!

You are right. Thanks for explaining

include/realtime_tools/realtime_box_best_effort.h
```diff
@@ -0,0 +1,232 @@
...
+template <class T, typename mutex_type = std::mutex>
+class RealtimeBoxBestEffort
+{
+  static_assert(
+    std::is_same_v<mutex_type, std::mutex> || std::is_same_v<mutex_type, std::recursive_mutex>);
+  static_assert(std::is_copy_constructible_v<T>, "Passed type must be copy constructible");
+
+public:
+  using mutex_t = mutex_type;
+  using type = T;
```
saikishor, line 64:
```suggestion
```

This is not being using, so I belive we can remove it

include/realtime_tools/realtime_box_best_effort.h
saikishor: Yes, but redefining another trait on top of same trait doesn't make sense right?, I think it make sense to use same `rcpputils::is_pointer<T>::value` without adding another one on top, if they are doing exactly the same job. 

What do you guys think about it? @bmagyar @destogl  @christophfroehlich

include/realtime_tools/realtime_box_best_effort.h
firesurfer: It is quite common to include this in containers -> So if a user builds some template magic around it, it is easier to access the underlying type (at least most stl containers do that).

include/realtime_tools/realtime_box_best_effort.h
firesurfer: I have to admit I like it this way as is makes clear that it works with pointers and smart pointers.

include/realtime_tools/realtime_box_best_effort.h
firesurfer: What we could do is use internally `type` instead of `T` at every occurance.

saikishor: Fine for me

firesurfer:
@saikishor 
If `constexpr auto is_ptr_or_smart_ptr = rcpputils::is_pointer<T>::value;` is the only thing keeping this back from getting  merged I can replace it. Would be defining an alias like `constexpr auto is_pointer_v = rcpputils::is_pointer<T>::value;` ok for you. I think it would keep the code more readable in general.

saikishor:
> @saikishor If `constexpr auto is_ptr_or_smart_ptr = rcpputils::is_pointer<T>::value;` is the only thing keeping this back from getting merged I can replace it. Would be defining an alias like `constexpr auto is_pointer_v = rcpputils::is_pointer<T>::value;` ok for you. I think it would keep the code more readable in general.

@firesurfer If others are okay with it, then it is fine and can be merged. For me, it is not with the naming `is_ptr_or_smart_ptr`, if it is the same trait doing the same why redefine/rename it and use it differently instead of using the original `rcpputils::is_pointer`.

This is just nitpicking, apart from that everything else looks good to me.

saikishor (approved):
Apart from the nitpicking discussion we are having, the logic and the idea of this PR look good to me.

If it is fine how it is, then I think it is good to be merged

firesurfer:
So if I understood it correctly we are just waiting for a comment on the `is_ptr_or_smart_ptr` alias from @christophfroehlich and @destogl  ?

saikishor:
> So if I understood it correctly we are just waiting for a comment on the `is_ptr_or_smart_ptr` alias from @christophfroehlich and @destogl ?

Hello @firesurfer!

As I have mentioned it is just a nitpicking comment, the changes in the PR look good to me. If they are okay with it, it can be merged.

christophfroehlich (approved):
Thanks for you contribution (and perseverance).

include/realtime_tools/realtime_box_best_effort.h
christophfroehlich: I don't have a clear opinion about this.

PR #140, opened 2024-01-13
Added method to set thread affinity for a certain core

firesurfer:
Hi,
this PR adds an additional method which allows "binding" the calling thread to a certain core. In fact it sets the thread affinity to a certain core. 

Even though @fmauch figured out in his tests (https://github.com/UniversalRobots/Universal_Robots_ROS_Driver/blob/60f08359f6/ur_robot_driver/doc/real_time_benchmarking.md) that setting the core affinity didn't change to much for his system, this can not be assumed in general. (See https://stackoverflow.com/questions/24645880/set-cpu-affinity-when-create-a-thread for a nice write up) Especially for more heterogene systems like the AMD Ryzen processes with 3D cache which is only connected to one processor die tying a thread to a certain core makes sense. 
Having the scheduler preferre a certain core for a certain thread can reduce cache misses and avoid context switches.


include/realtime_tools/thread_priority.hpp
```diff
@@ -46,6 +46,18 @@ bool has_realtime_kernel();
  */
 bool configure_sched_fifo(int priority);
 
+/**
+ * Configure the caller thread affinity - Tell the scheduler to prefer a certain core for the current thread
+ * \param[in] core - the number of the core
+ * \returns true if configuring the scheduler succeeded
+*/
+bool set_preferred_core(const int core = -1);
```
saikishor, line 54:
```suggestion
/**
 * Configure the caller thread affinity - Tell the scheduler to prefer a certain core for the current thread
 * \param[in] core - the number of the core
 * When parsed -1, the CPU affinity is reset to all CPUs
 * \returns true if configuring the scheduler succeeded
*/
bool set_cpu_affinity(const int core);
```

I would rather make the naming more standard as known to everyone

include/realtime_tools/thread_priority.hpp
```diff
@@ -46,6 +46,18 @@ bool has_realtime_kernel();
...
+
+/**
+ * \returns The amount of available cpu cores
+*/
+int get_core_count();
```
saikishor, line 59:
```suggestion
int get_available_cpu_count();
```

saikishor (changes requested):
I think something like this would be good

If you can add support to [Windows](https://stackoverflow.com/questions/150355/programmatically-find-the-number-of-cores-on-a-machine) or disable for WINDOWS

PR #146, opened 2024-01-25
Replace existing RealtimeBox implementation with RealtimeBoxBestEffort implementation

firesurfer:
Hi as said in #139 this PR replaces the existing RealtimeBox implementation with the new implementation provided in #139 . I merged the tests into a single file and they work fine (even though the existing implementation had only a very small amount of tests)

The implementation of the get/set methods matches exactly the one from the existing implementation apart from some template magic which disables the standard get/set methods for pointer types. (For which they wouldnt provide any thread safety btw from my point of view - see #138) 

TLDR: Merging this has the potential to __break existing code__ if someone used pointer types with the RealtimeBox. For everything else it should work just fine. 

I would also be happy to provide some documentation on how to use this implementation but I am not sure where in the repository it should go. 

This PR should be merged after #139 

EDIT: One question regarding the copyright notice for the existing tests. How should this be handled? At the moment I didn't copy any of it to the merged file.


firesurfer:
@christophfroehlich I will update this PR as soon as #139 is finished

firesurfer:
@christophfroehlich Given that #139 is now merged I updated this PR. 
Please see my notes in the initial description about open points for discussion.

Also pinging @saikishor

firesurfer:
@christophfroehlich Just a note about the CI. Interestingly I just had the case that pre-commit in the CI failed but did run sucessfully on my system. (Note about using the right include guard - REALTIME_TOOLS__REALTIME_BOX_H_)

christophfroehlich:
> @christophfroehlich Just a note about the CI. Interestingly I just had the case that pre-commit in the CI failed but did run sucessfully on my system. (Note about using the right include guard - REALTIME_TOOLS__REALTIME_BOX_H_)

Strange. Have you run `pre-commit run --all`? otherwise it checks only the changed files in the index.

firesurfer:
> Strange. Have you run pre-commit run --all? otherwise it checks only the changed files in the index.

Yes that's what I did run. But I couldn't reproduce it now. .

christophfroehlich:
> I would also be happy to provide some documentation on how to use this implementation but I am not sure where in the repository it should go.

We don't have a RST documentation of this repo, but I plan to rework the API documentation somewhen on control.ros.org. For now, please just add anything in the manner of [rosdoc2](https://github.com/ros-infrastructure/rosdoc2) to be published on [index.ros.org](https://index.ros.org/p/realtime_tools/github-ros-controls-realtime_tools/#rolling)

saikishor:
Is there a way to add some methods with old pointers and deprecation warnings?

firesurfer:
The current implementation in the RealTimeBoxBestEffort basically disables the normal `trySet, tryGet...etc` methods for pointer types. 

Given that it is inherently unsafe to use pointers with this implemenation and the default `trySet,tryGet` methods (and also the existing RealTimeBox implementation) one could argue that it could be profitable to actually break builds here. 

Nevertheless I completely understand your argument. 

I guess adding a deprecation notice might be possible but I am not sure how at the moment. We could use SFINEA to provide a  version of the class for pointers that include the deprecation notices (and removes the `typename std::enable_if_t<!is_ptr_or_smart_ptr<U>, U> ` for the normal `trySet, tryGet` methods). 



In the end I think this really comes to the question: Allow wrong behavior or possibly break builds?

saikishor:
> The current implementation in the RealTimeBoxBestEffort basically disables the normal `trySet, tryGet...etc` methods for pointer types.
> 
> Given that it is inherently unsafe to use pointers with this implemenation and the default `trySet,tryGet` methods (and also the existing RealTimeBox implementation) one could argue that it could be profitable to actually break builds here.
> 
> Nevertheless I completely understand your argument.
> 
> I guess adding a deprecation notice might be possible but I am not sure what is the best way at the moment. We could use SFINEA to provide a version of the class for pointers that include the deprecation notices (and removes the `typename std::enable_if_t<!is_ptr_or_smart_ptr<U>, U> ` for the normal `trySet, tryGet` methods).
> 
> In the end I think this really comes to the question: Allow wrong behavior or possibly break builds?

I already agree with you on the unsafe part from the other PR, the problem here is we have a single branch that runs for Rolling, Iron, and also Humble, so breaking it would be breaking the setup of a lot of users, and such API breaking is usually not seen good. 

I could think of the following 3 options: 

* Branch the rolling from others and then make this `RealTimeBoxBestEffort` as default over there.
* Add a deprecation notice on the current one and ask users to move to the updated one this deprecation can be completely removed after one ROS release, so the users can migrate gradually. 
* We can somehow try to wrap the raw pointer of the user to a smart pointer and then use the current class. 

What do you think about this?

firesurfer:
> Branch the rolling from others and then make this RealTimeBoxBestEffort as default over there.

This introduces additional maintenance effort -> I can not judge if this would be worth the effort

>    Add a deprecation notice on the current one and ask users to move to the updated one this deprecation can be completely removed after one ROS release, so the users can migrate gradually.

According to https://stackoverflow.com/questions/48045559/how-do-i-declare-sfinae-class I could do that.
The idea would be: Provide a specialisation for pointers, enable the methods there and add deprecation notes to them

I am currently trying to figure out a way to do this

>    We can somehow try to wrap the raw pointer of the user to a smart pointer and then use the current class.

This does not solve the problem. As the aim of this implementation is to provide thread safe access to its contents the access to any pointer type has to be wrapped in the way shown in the examples. (Using the shared_ptr does not make it safer than using a raw pointer)

firesurfer:
@saikishor I just pushed a version which add a specialisation for pointer types. The tests will now show a deprecation note if get/set are used with pointer types. Also the code starts looking like a stl class...not sure if thats a good thing :D

christophfroehlich:
> EDIT: One question regarding the copyright notice for the existing tests. How should this be handled? At the moment I didn't copy any of it to the merged file.

Please just add another line to the copyright claim and don't remove the old one if old code remained inside (at least I saw this in other ROS repos)

firesurfer:
@saikishor Perhaps you'll find some time for another review? It will now handle pointers types as before but print a deprecation warning

firesurfer:
Just a friendly ping :)

include/realtime_tools/realtime_box.h
```diff
@@ -26,50 +26,402 @@
...
+    }
+    value_ = value;
+    return true;
+  }
+  /**
+   * @brief access the content readable with best effort
+   * @return false if the mutex could not be locked
+   * @note only safe way to access pointer type content (rw)
+   */
+  bool trySet(const std::function<void(T &)> & func)
```
bmagyar, line 107:
```suggestion
  bool try_set(const std::function<void(T &)> & func)
```
we've been following snake-case in ros2_control, please adjust the rest too

firesurfer: To be honest I am a bit puzzled about this comment given that the PR that introduced these names was accepted https://github.com/ros-controls/realtime_tools/pull/139

This PR ist just for replacing the old RealtimeBox with the new implementation while still keeping it backward compatible with the corresponding deprecation warnings.

include/realtime_tools/realtime_box.h
```diff
@@ -26,50 +26,402 @@
...
+
+/**
+ * @brief Specialisation for pointer types.
+ * WHY is this specialised. We do not want to break compatibility but show a deprecation note
+ * for get/set etc. methods if used with a pointer type
+ * They are unsafe to use and should therefore be replace with their correspondents that take an std::function for accessing
+ * the value behind the pointer
+*/
+template <class T, typename mutex_type>
+class RealtimeBox<T, mutex_type, true>
```
saikishor, line 248:
```suggestion
template <class T, typename mutex_type>
[[deprecated("RealtimeBox is not recommended to use with Raw pointers! Use smart pointers instead.")]]
class RealtimeBox<T, mutex_type, true>
```

saikishor, line 248: @firesurfer In this overlayed class, do we need to have a copy of most of the methods similar to the previous one? or is there a way to reuse them?

test/realtime_box_tests.cpp
```diff
@@ -29,36 +30,204 @@
...
+}
+
+TEST(RealtimeBox, pointer_type)
+{
+  int a = 100;
+  int * ptr = &a;
+
+  RealtimeBox<int *> box(ptr);
+  // This does not and should not compile!
+  // auto value = box.get();
```
saikishor, line 146:
```suggestion
```

test/realtime_box_tests.cpp
```diff
@@ -29,36 +30,204 @@
...
+  box.tryGet([](const auto & i) { EXPECT_EQ(*i, 200); });
+}
+
+TEST(RealtimeBox, smart_ptr_type)
+{
+  std::shared_ptr<int> ptr = std::make_shared<int>(100);
+
+  RealtimeBox<std::shared_ptr<int>> box(ptr);
+  // This does not and should not compile!
+  // auto value = box.get();
```
saikishor, line 165:
```suggestion
```

saikishor (changes requested):
@firesurfer few minor comments. 

Thank you and sorry for the huge delay

include/realtime_tools/realtime_box.h
firesurfer: @saikishor From the "base" implementation:

> // Provide a specialisation for non pointer types
>  // NOTE: When migrating to a safe access only version just remove the specialisation for pointer
>  // and let this be the only version!

My argument would be that in the longterm the pointer specialisation should be removed. I think the only save way to access the values behind pointers is to use the "get/set(std::function) overloads). 
By keeping the code duplicated we can simplify this process. 

My suggestion is to keep this version for at least one release and remove the pointer specialisation in the next one.

firesurfer: It should actually be: Only use the "get/set(std::function) overloads of the RealtimeBox (which is the reason why we added this template specialisation for pointers).

But that deprecation message is already printed when the default get/set methods are used with pointers!

test/realtime_box_tests.cpp
firesurfer: At the moment this would compile btw. (Because we added the pointer specialisation for compatibility reasons)

saikishor: Sure, then let's remove the comment.

test/realtime_box_tests.cpp
saikishor: Sure, then let's remove the comment.

include/realtime_tools/realtime_box.h
saikishor: Hmm! @bmagyar @christophfroehlich any strong opinion on this one?

firesurfer: Perhaps I should give a quick recap why it is designed like this given that quite some time has passed now:

1. The existing RealtimeBox implementation is not safe to use with any kind of pointer. Because only any operation on the pointer itself would be locked by the mutex but not the access to the data behind the pointer 

2. The RealtimeBoxBestEffort solves this issue by providing additional overloads to get/set which take a function. These overloads lock the mutex and call the function. Inside this function manipulating the pointer is safe. 
Additionally if the Type is a pointer the normal get/set methods which return a reference to the internal value are disabled via SFINEA (std::enable_if)

3. It makes sense to replace the existing RealtimeBox with the RealtimeBoxBestEffort (This is what this PR aims for). Additionally the requirement was made: We shall not break existing code, but we want to display deprecation notices if the default get/set methods are used with pointers. Therefore I added a template specialisation for pointers removes the std::enable_if for get/set with pointer types and adds deprecation notices for the get/set methods without the std::function overloads. 

This should encourage users to migrate to the get/set(std::function) version for pointer types. 
So in a future release the pointer specialisation of the class can be removed. 

So to answer your question:

Pointer types are supported BUT you need to use the get/set(std::function) methods.

saikishor: @firesurfer I understand your point. What I'm saying is, we don't have to break the Raw pointer in this version, but in the future versions, we can make the users use the smart pointers. So, if we mark the class deprecated with the notice to use the smart pointers in future releases, the users will be able to migrate in the meantime. 

If we mark the methods alone, they still think the Raw pointers will be supported in the future. I'm not sure, if it is interesting to support Raw pointer in the future. If we still want to support, then the way you marked the depreciation is good enough @bmagyar @christophfroehlich any comments on this?

christophfroehlich:
hi @firesurfer! after the big changes to ros2_control are done, we now really want to get this PR merged soon.

- Could you please add https://github.com/ros-controls/realtime_tools/pull/183 to this PR
- and address the snake_case as @bmagyar mentioned? (yes, we should have persisted on that when the first realtime box PR was merged.) As you might be the only one using it until now, you can decide if we should go with the API break due to the renaming or add wrappers with deprecation notes.

If we manage to merge this within next week, it might go into the final iron release as well.

firesurfer:
@christophfroehlich I'll try to find sometime to work on this this week

firesurfer:
@christophfroehlich @bmagyar 
I removed the pointer deprecation overload (which makes this class way more maintainable)  and made the method names all snake_case.

firesurfer:
I just realized that I forgot to add copy and move constructors (needed due to the std::mutex not being copyable/movable) and added them.

saikishor (changes requested):
@firesurfer sorry for not getting back sooner. Do you think it is possible to add deprecation notes to the older methods? I think it's better to have a smoother transition to others. What do you think?

BTW do you know which packages we will break with not supporting raw pointers anymore?

firesurfer:
@saikishor Could you please elaborate what you mean with "older methods" and by "pages" ?

saikishor:
> @saikishor Could you please elaborate what you mean with "older methods" and by "pages" ?

@firesurfer I mean the CamelCase methods that you removed in the recent commits. Can we add a deprecation notice to them?

"Pages" is a typo here, I just fixed it. I was referring to the packages. Do you know which packages we will break with not supporting raw pointers anymore?

christophfroehlich:
> "Pages" is a typo here, I just fixed it. I was referring to the packages. Do you know which packages we will break with not supporting raw pointers anymore?

afaik (according to index.ros.org) there is no package listing realtime_tools as a direct dependency using the RealtimeBox. Apart from that it is hard to track dependencies if they aren't released.

firesurfer:
@saikishor I don't think that is necessary. They were added just recently with the RealtimeBoxBestEffort and I think nobody is using them yet.

Regarding the second question:
I did a quick search on github: https://github.com/search?q=RealtimeBox%3C&type=code
But as @christophfroehlich said: hard to tell: I looked into some of those packages and they were still ROS1.

christophfroehlich:
- [ ]   ros-rolling-ur-controllers
- [ ]   ros-rolling-picknik-twist-controller
- [ ]   ros-rolling-picknik-reset-fault-controller
- [ ]   ros-rolling-moveit-servo
- [ ]   ros-rolling-canopen-ros2-controllers
- [ ]   ros-rolling-battery-state-broadcaster
- [ ]   ros-jazzy-ur-controllers
- [ ]   ros-jazzy-picknik-twist-controller
- [ ]   ros-jazzy-picknik-reset-fault-controller
- [ ]   ros-jazzy-moveit-servo
- [ ]   ros-jazzy-battery-state-broadcaster
- [ ]   ros-humble-ur-controllers
- [ ]   ros-humble-picknik-twist-controller
- [ ]   ros-humble-picknik-reset-fault-controller
- [ ]   ros-humble-clearpath-mecanum-drive-controller
- [ ]   ros-humble-canopen-ros2-controllers

christophfroehlich:
The only package of the listed dependencies above which use any of the RealtimeBox/RealTimeBoxBestEffort classes is this ur_controller: It uses a shared pointer, but will fail if we delete the RealTimeBoxBestEffort classname. Would a `using RealTimeBoxBestEffort=RealTimeBox;` work here @firesurfer? :eyes: @fmauch 
https://github.com/UniversalRobots/Universal_Robots_ROS2_Driver/blob/main/ur_controllers/include/ur_controllers/ur_configuration_controller.hpp#L91-L93

firesurfer:
Our @fmauch might want to fix this ? I think this controller is rather new - so the impact of fixing this would be managable. 


@saikishor 

> I could see these released packages are using it in rolling, we just have to make sure that all of these are using shared_ptrs instead of raw pointers. We will have to check the same for Jazzy and Humble

From my understanding using a shared_ptr isn't safer in anyway (apart from memory safety) than a normal pointer with the RealTimeBox -> That is the reason I enforced using the methods that are overloaded with a `std::function` for normal pointers and shared_ptrs. 

The discussion is basically the same as in the beginning: Shall we enforce correct behavior and possibly break dependent code or do we add back the deprecation workaround - which is annoying to maintain.

saikishor:
> And in order to answer your question @christophfroehlich - `using RealTimeBoxBestEffort=RealTimeBox;` should work

Yup let's do that. I don't know if we can add deprecation notice to the using keyword. If so, we can add it and ask everyone to switch to new RealTimeBox?

> @saikishor 
> 
> From my understanding using a shared_ptr isn't safer in anyway (apart from memory safety) than a normal pointer with the RealTimeBox -> That is the reason I enforced using the methods that are overloaded with a `std::function` for normal pointers and shared_ptrs. 

You are absolutely right. The other day I was checking the same and it doesn't make sense to use Pointers here, as you can still change it outside the RealTimeBox class. As you said, it is meant for classes and structs to be used safely.


> The discussion is basically the same as in the beginning: Shall we enforce correct behavior and possibly break dependent code or do we add back the deprecation workaround - which is annoying to maintain. 

As the Original API is simple and it already mentions in the documentation note to use shared_ptrs I don't mind merging it and see what will be the case.

I apologize for delaying this so far.

saikishor:
@firesurfer I'll do one final review to see if I find something.

include/realtime_tools/realtime_box.h
```diff
@@ -26,50 +26,251 @@
...
- */
-template <class T>
+// Provide a specialisation for non pointer types
+// NOTE: When migrating to a safe access only version just remove the specialisation for pointer
+// and let this be the only version!
+template <class T, typename mutex_type = std::mutex>
 class RealtimeBox
 {
+  static_assert(
+    std::is_same_v<mutex_type, std::mutex> || std::is_same_v<mutex_type, std::recursive_mutex>);
```
saikishor, line 64: @firesurfer do we have a need for recursive mutex here?, may be it is related to the `get_mutex` methods you implemented.

If so, can we rename this class to `RealTimeBoxBase` and with `using` keyword we create instances of different mutexes. The cool thing with that is when we have the priority inheritance mutexes, we can simply replace there and everyone using that code will have it propagated. 

What do you think?

saikishor (commented):
@firesurfer Overall looks good to me, I've opened discussion at some points.

include/realtime_tools/realtime_box.h
```diff
@@ -1,4 +1,4 @@
-// Copyright (c) 2009, Willow Garage, Inc.
+// Copyright (c) 2024, Lennart Nachtigall
```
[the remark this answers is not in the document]
firesurfer: Let's add both. This was due to way it was merged.

include/realtime_tools/realtime_box.h
```diff
@@ -26,50 +26,251 @@
...
+
 private:
-  // The thing that's in the box.
-  T thing_;
-
-  // Protects access to the thing in the box.  This mutex is
-  // guaranteed to be locked for no longer than the duration of the
-  // copy, so as long as the copy is realtime safe and the OS has
-  // priority inheritance for mutexes, this lock can be safely locked
-  // from within realtime.
```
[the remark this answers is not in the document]
firesurfer: Yes, but I am not sure about the copy. It is possible to use this box without actually making any copies :D

include/realtime_tools/realtime_box.h
firesurfer: Well this class already allows to template the type of mutex. I thought it might be a good idea to limit which mutexes are allowed. But we can remove this static_assert

include/realtime_tools/realtime_box.h
```diff
@@ -26,50 +26,251 @@
...
+    return get();
+  }
+  /**
+   * @brief provide a custom conversion operator
+   * @note Can be used from non-RT and RT contexts
+   */
+  template <typename U = T, typename = typename std::enable_if_t<!is_ptr_or_smart_ptr<U>>>
+  [[nodiscard]] operator std::optional<T>() const
+  {
+    return try_set();
```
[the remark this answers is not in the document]
firesurfer: Good that you spotted that !

include/realtime_tools/realtime_box.h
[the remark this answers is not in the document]
firesurfer: I leave it up to you to. I don't have any strong opinion about it.

include/realtime_tools/realtime_box.h
```diff
@@ -26,50 +26,251 @@
...
+  }
+
+  // In case one wants to actually use a pointer
+  // in this implementation we allow accessing the lock directly.
+  // Note: Be careful with lock.unlock().
+  // It may only be called from the thread that locked the mutex!
+  [[nodiscard]] const mutex_t & get_mutex() const { return lock_; }
+  [[nodiscard]] mutex_t & get_mutex() { return lock_; }
```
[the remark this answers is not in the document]
saikishor: Sounds fair to me!

include/realtime_tools/realtime_box.h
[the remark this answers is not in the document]
saikishor: Let's put it back, I'll remove it once we add priority inheritance mutexes

include/realtime_tools/realtime_box.h
saikishor: > If so, can we rename this class to RealTimeBoxBase and with `using` keyword we create instances of different mutexes. 

What do you think about this approach? And have things with using? The default RealTimeBox instance will use std::mutex as here

firesurfer: I just wanted to understand the why :)

I can add them, as well as a using for the RealTimeBoxBestEffort with a deprecation notice.

saikishor: If you are convinced, let's do it. If not, not a problem 😉

firesurfer: So I added them. Please take a look if you agree with the names....naming things is always hard :D 

And @fmauch should now get a nice deprecation warning for the RealTimeBoxBestEffort

include/realtime_tools/realtime_box.h
```diff
@@ -26,52 +27,270 @@
...
+// Provide specialisations for different mutex types
+template <typename T>
+using RealtimeBoxStandard = RealtimeBox<T, std::mutex>;
+
+template <typename T>
+using RealtimeBoxRecursive = RealtimeBox<T, std::recursive_mutex>;
+
+// This is the specialisation we recommend to use in the end
+template <typename T>
+using RealtimeBox = RealtimeBoxDefault<T>;
```
saikishor, line 292: Which is the default one? I don't see it's definition in this file

include/realtime_tools/realtime_box.h
```diff
@@ -26,52 +27,270 @@
...
 
+// Introduce some easier to use names
+
+// Only kept for compatibility reasons
+template <typename T, typename mutex_type = std::mutex>
+using RealtimeBoxBestEffort [[deprecated]] = RealtimeBox<T, mutex_type>;
+
+// Provide specialisations for different mutex types
+template <typename T>
+using RealtimeBoxStandard = RealtimeBox<T, std::mutex>;
```
saikishor, line 285: Why not just call it just `RealtimeBox`?

include/realtime_tools/realtime_box.h
firesurfer: Sorry I just messed up a little bit. Can you take a look at the latest commit ?

saikishor: Sure

include/realtime_tools/realtime_box.h
```diff
@@ -26,52 +27,270 @@
...
   // from within realtime.
-  std::mutex thing_lock_RT_;
+  mutable mutex_t lock_;
 };
 
+// Introduce some easier to use names
+
+// Only kept for compatibility reasons
+template <typename T, typename mutex_type = std::mutex>
+using RealtimeBoxBestEffort [[deprecated]] = RealtimeBoxBase<T, mutex_type>;
```
saikishor, line 281:
```suggestion
using RealtimeBoxBestEffort [[deprecated ("Use RealtimeBox instead")]] = RealtimeBoxBase<T, mutex_type>;
```

saikishor (changes requested):

include/realtime_tools/realtime_box.h
saikishor: Looks good to me

fmauch:
> Our @fmauch might want to fix this ? I think this controller is rather new - so the impact of fixing this would be managable.

I'll add it to my tech debt list :-) Thanks for adding the deprecation layer, though!

include/realtime_tools/realtime_box.h
```diff
@@ -26,52 +27,271 @@
 // ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
 // POSSIBILITY OF SUCH DAMAGE.
 
-// Author: Stuart Glaser
+// Author: Lennart Nachtigall
```
saikishor, line 30:
```suggestion
// Author: Stuart Glaser
// Author: Lennart Nachtigall
```

christophfroehlich (approved):
LGTM as far as I can assess this. Thanks again for the follow-up on this

include/realtime_tools/realtime_box.h
```diff
@@ -27,51 +28,271 @@
...
+    // Lock the other box mutex
+    std::unique_lock<mutex_t> lock(o.lock_);
+    // We do not need to lock our own mutex because we are currently in the process of being created
+    value_ = std::move(o.value_);
   }
+  // Only enabled for types that can be constructed from an initializer list
+  template <typename U = T>
+  constexpr RealtimeBoxBase(
+    const std::initializer_list<U> & init,
+    std::enable_if_t<std::is_constructible_v<U, std::initializer_list>>)
```
christophfroehlich, line 102:
```suggestion
    std::enable_if_t<std::is_constructible_v<U, std::initializer_list<U>>>)
```
is this correct, see #183?

christophfroehlich (changes requested):
see the clang fix above

christophfroehlich (approved):

PR #151, opened 2024-02-11
[CI] Use reusable workflows and matrix strategy

christophfroehlich:
The git diff is broken here unfortunately.

- I used the matrix strategy to reduce the number of workflow files (humble, iron, rolling run from the same branch)
- Use reusable workflow files from https://github.com/ros-controls/ros2_control_ci
- Use reviewer database from ros2_control_ci, see https://github.com/ros-controls/ros2_control/pull/1402


bmagyar (approved):

fmauch (approved):
Looks good, just two minor comments:

- the `ref_for_scheduled_build` isn't strictly needed, since everything runs from the master branch, anyway, but it might be good to keep as a reminder
- I do like the one matrix to rule them all approach, it's just sad that we cannot get individual badges per matrix entry... But especially since everything runs from the master branch, using one common workflow file rules out that disadvantage in my opinion.

---

Edit: Too slow...

christophfroehlich:
> Looks good, just two minor comments:
> 
>     * the `ref_for_scheduled_build` isn't strictly needed, since everything runs from the master branch, anyway, but it might be good to keep as a reminder
> 
>     * I do like the one matrix to rule them all approach, it's just sad that we cannot get individual badges per matrix entry... But especially since everything runs from the master branch, using one common workflow file rules out that disadvantage in my opinion.

I agree with both. It's too easy to forget something when updating the individual workflows, that's why I vote for the matrix one.

PR #161, opened 2024-04-08
Add custom rosdoc2 config

christophfroehlich:
rosdoc2 will be used by the build farm to publish API docs on http://docs.ros.org/en/rolling/p/realtime_tools/

- Add custom landing page for rosdoc2: logo, links to control.ros.org
- Copy over information from https://wiki.ros.org/realtime_tools
- add rosdoc2 workflow, but only if the relevant files changed

image


.github/workflows/ci-pre-commit.yml
```diff
@@ -8,7 +8,7 @@ on:
 
 jobs:
   pre-commit:
-    uses: ros-controls/ros2_control_ci/.github/workflows/reusable-pre-commit.yml@master
+    uses: ros-controls/ros2_control_ci/.github/workflows/reusable-pre-commit.yml@doxygen
```
christophfroehlich, line 11:
```suggestion
    uses: ros-controls/ros2_control_ci/.github/workflows/reusable-pre-commit.yml@master
```

doc/index.rst
```diff
@@ -0,0 +1,59 @@
+Welcome to the documentation for realtime_tools
+===============================================
+
+Contains a set of tools that can be used from a hard realtime thread, without breaking the realtime behavior.
+
+For more information of the ros2_control framework see `control.ros.org <https://control.ros.org/>`__.
+
+Realtime Publisher
+------------------
+The ``realtime_tools::RealtimePublisher`` allows users that write C++ ros2_controllers to publish messages on a ROS topic from a hard realtime loop. The normal ROS publisher is not realtime safe, and should not be used from within the update loop of a realtime controller. The realtime publisher is a wrapper around the ROS publisher; the wrapper creates an extra non-realtime thread that publishes messages on a ROS topic. The example below shows a typical usage of the realtime publisher in the ``on_configure()`` and ``update()`` methods of a realtime controller:
```
fmauch, line 10: I think in the context of this example it would be worth mentioning that `on_configure()` is not a realtime method. For somebody not knowing the framework this might not be clear and creating the publisher is clearly not a realtime-compatible call.

```suggestion
The ``realtime_tools::RealtimePublisher`` allows users that write C++ ros2_controllers to publish messages on a ROS topic from a hard realtime loop. The normal ROS publisher is not realtime safe, and should not be used from within the update loop of a realtime controller. The realtime publisher is a wrapper around the ROS publisher; the wrapper creates an extra non-realtime thread that publishes messages on a ROS topic. The example below shows a typical usage of the realtime publisher in the ``on_configure()`` (non-realtime method) and ``update()`` (realtime method) methods of a realtime controller:
```

fmauch (commented):
+1 for per-package documentation :-)

doc/index.rst
christophfroehlich: Thanks. Honestly, I just copied that from wiki.ros.org and updated it for ROS 2. But you are totally right here.

bmagyar (approved):

PR #164, opened 2024-04-24
[CI] Use matrix strategy

christophfroehlich:
as per request of @fmauch


fmauch (approved):
> as per request of @fmauch

It was more of a suggestion ;-)

christophfroehlich:
included in #165

PR #172, opened 2024-09-13
[AsyncFunctionHandler] Add exception handling

saikishor:
Right now, the async function handler doesn't work well when the exceptions are then in the callback and this might be an issue that we might face in the short term.

This PR addresses this issue by handling the exceptions and throw in the thread that calls `trigger_async_callback`, so that the main thread is aware of the exception and handle it properly


destogl (approved):
Great and clean.

bmagyar (approved):

PR #173, opened 2024-09-30
Include <functional> in realtime_box_best_effort

fmauch:
Since the `RealTimeBoxBestEffort` uses `std::function`s to set and get pointer types (which is great), it should also include `<functional>`.

Note: This doesn't explode in the tests, since there is `gmock` included, which somewhere includes <functional> on the way.


saikishor (approved):
LGTM

christophfroehlich (approved):

PR #175, opened 2024-10-28
Add a helper method to lock the pages of memory in the RAM

saikishor:


include/realtime_tools/thread_priority.hpp
```diff
@@ -44,6 +46,13 @@ bool has_realtime_kernel();
  */
 bool configure_sched_fifo(int priority);
 
+/**
+ * Locks the memory pages of the calling thread to prevent page faults. By calling this method, the programs locks all pages mapped into the address space of the calling process and future mappings. This means that the kernel will not swap out the pages to disk i.e., the pages are guaranteed to stay in RAM until later unlocked - which is important for realtime applications.
+ * \param[out] message a message describing the result of the operation
+ * \returns true if memory locking succeeded, false otherwise
+*/
+bool lock_memory(std::string & message);
```
aprotyas, line 54: Seems like a candidate to adopt `std::expected<std::monostate, std::string>`

aprotyas (approved):

src/thread_priority.cpp
```diff
@@ -53,4 +55,49 @@ bool configure_sched_fifo(int priority)
...
+
+bool lock_memory(std::string & message)
+{
+  if (mlockall(MCL_CURRENT | MCL_FUTURE) == -1) {
+    if (!is_capable(CAP_IPC_LOCK)) {
+      message = "No proper privileges to lock the memory!";
+    } else if (errno == ENOMEM) {
+      message =
+        "The caller had a nonzero RLIMIT_MEMLOCK soft resource limit, but tried to lock more "
+        "memory than the limit permitted. Try running application with privileges!";
```
fmauch, line 84:
```suggestion
        "memory than the limit permitted. Try running the application with privileges!";
```

src/thread_priority.cpp
```diff
@@ -53,4 +55,49 @@ bool configure_sched_fifo(int priority)
...
+bool is_capable(cap_value_t v)
+{
+  bool rc = false;
+  cap_t caps;
+  if ((caps = cap_get_proc()) == NULL) {
+    return false;
+  }
+
+  if (cap_set_flag(caps, CAP_EFFECTIVE, 1, &v, CAP_SET) == -1) {
+    goto cleanup;
```
fmauch, line 67: Instead of a `goto`: Could we use a lambda function there?

saikishor: Done

saikishor:
I'm also thinking if it makes sense to continue to `thread_priority.hpp` or something like `realtime_helpers.hpp` as it now contains the part of memory locking etc and in the future may be more. So, the header would have a meaningful name. We can add a backward compatibility with deprecation note, if we are interested to do this

include/realtime_tools/thread_priority.hpp
aprotyas: Oh, true. I was just musing anyway.

saikishor: No worries, it is interesting to learn this alternative for the future. Thank you

fmauch (approved):
Looks good to me.

I like the proposal regarding renaming this, but maybe this should be a separate PR? This way we can backport this without breaking the API.

christophfroehlich (approved):
debian an RHEL are happy now, so am I

PR #181, opened 2024-10-31
[AsyncFunctionHandler] return execution time in nanoseconds

saikishor:
I think this is useful to have more statistical data when we add to the CM

### DIFF
--- a/include/realtime_tools/async_function_handler.hpp
+++ b/include/realtime_tools/async_function_handler.hpp
@@ -169,7 +169,7 @@ public:
     std::unique_lock<std::mutex> lock(async_mtx_);
     stop_async_callback_ = false;
     trigger_in_progress_ = false;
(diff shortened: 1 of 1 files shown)


bmagyar (commented):
Ah this is nice! Thank you

christophfroehlich (commented):
Shouldn't a nanosecond be represented in a double value of seconds? The time delta will be <1.0 -> we have ~15 digits after the comma?

saikishor:
> Shouldn't a nanosecond be represented in a double value of seconds? The time delta will be <1.0 -> we have ~15 digits after the comma?

@christophfroehlich 
Yes, but if you want to see the spikes in that 15 digits, it is not very easy. Instead, having it as nanoseconds can let you easy. This is the most granularity you can get

christophfroehlich (approved):

PR #186, opened 2024-11-03
Add methods to set the thread affinity

saikishor:
This PR uses the commits from PR: #140 and add more fixes on top of it and some refactoring and renaming of methods

Closes  #140

### DIFF
--- a/src/realtime_helpers.cpp
+++ b/src/realtime_helpers.cpp
@@ -116,7 +116,7 @@ bool lock_memory(std::string & message)
 #endif
 }
 
-bool set_thread_affinity(int core, int pid, std::string & message)
+bool set_thread_affinity(int pid, int core, std::string & message)
 {
 #ifdef _WIN32
   message = "Thread affinity is not supported on Windows.";
--- a/src/realtime_helpers.cpp
+++ b/src/realtime_helpers.cpp
@@ -135,7 +135,7 @@ static void print_error(const int errc)
   }
 }
 
-bool set_cpu_affinity(int core)
+bool set_thread_affinity(int core)
 {
   // Allow attaching the thread/process to a certain cpu core
   cpu_set_t cpuset;
--- a/src/realtime_helpers.cpp
+++ b/src/realtime_helpers.cpp
@@ -118,6 +118,10 @@ bool lock_memory(std::string & message)
 
 bool set_thread_affinity(int core, int pid)
 {
+#ifdef _WIN32
+  std::cerr << "Thread affinity is not supported on Windows." << std::endl;
+  return false;
+#else
   auto print_error = [](int result) {
     if (result == 0) {
       return;
@@ -168,6 +172,7 @@ bool set_thread_affinity(int core, int pid)
   }
   // Invalid core number passed
   return false;
+#endif
 }
 
 int get_number_of_available_processors()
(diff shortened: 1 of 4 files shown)


christophfroehlich:
:eyes: @firesurfer do you support the changes?
@SENAI-GilmarCorreia does this work for you on your windows setup?

christophfroehlich (commented):
👀 @firesurfer do you support the changes?
@SENAI-GilmarCorreia does this work for you on your windows setup?

SENAI-GilmarCorreia:
@christophfroehlich, I'm having trouble with the #warning on line 35 in thread_priority.hpp. I'm submitting a PR.

saikishor:
> @christophfroehlich, I'm having trouble with the #warning on line 35 in thread_priority.hpp. I'm submitting a PR.
> 

@SENAI-GilmarCorreia the issue with the thread_priority.hpp is with this PR? Or with master branch?

firesurfer:
I think the `message` parameter is a bit weird API wise. 

As an alternative one could either:

a) Return a std::pair<bool, std::string>
b) Make passing the message parameter optional. I am not sure if in C++17 something like this actually works:

`bool set_cpu_affinity(int pid,int core, std::string& message = {}` 

In addition I would make the `pid` parameter optional or provide an additional function: `set_current_thread_affinity`

saikishor:
> I think the `message` parameter is a bit weird API wise. 
> 
> As an alternative one could either:
> 
> a) Return a std::pair<bool, std::string>
> b) Make passing the message parameter optional. I am not sure if in C++17 something like this actually works:
> 
> `bool set_cpu_affinity(int pid,int core, std::string& message = {}` 
> 
> In addition I would make the `pid` parameter optional or provide an additional function: `set_current_thread_affinity`
> 
> Other than that: Looks fine to me.

@firesurfer first of all, thank you for all your contributions 

Sure, I initially thought of the pair option but I skipped. The message parameter is nice, if you want to have the logging at a higher level, instead of directing it to the std::cerr. 

I can add an additional method called `set_current_thread_affinity`, it makes sense.

saikishor:
@firesurfer can you take a look now?. If it looks good to you, please approve the PR

src/realtime_helpers.cpp
```diff
@@ -34,10 +34,13 @@
 #include <sched.h>
 #include <sys/capability.h>
 #include <sys/mman.h>
+
+#include <unistd.h>
 #endif
 
 #include <cstring>
 #include <fstream>
+#include <iostream>
```
firesurfer, line 43: I think we do not need this include

saikishor: Done!

src/realtime_helpers.cpp
```diff
@@ -112,4 +115,86 @@ bool lock_memory(std::string & message)
...
+  CPU_ZERO(&cpuset);
+  // Obtain available processors
+  const auto number_of_cores = get_number_of_available_processors();
+
+  // Reset affinity by setting it to all cores
+  if (core < 0) {
+    for (auto i = 0; i < number_of_cores; i++) {
+      CPU_SET(i, &cpuset);
+    }
+    // And actually tell the schedular to set the affinity of the currently calling thread
```
firesurfer, line 160: The comment should be updated

saikishor: Done! Updated

firesurfer (approved):

SENAI-GilmarCorreia:
> @SENAI-GilmarCorreia the issue with the thread_priority.hpp is with this PR? Or with master branch?

@saikishor You are right! I will open a PR directly to the master branch.

SENAI-GilmarCorreia (approved):

christophfroehlich (approved):

PR #189, opened 2024-11-04
Fix libcap-dev dependency

saikishor:
@christophfroehlich reported the missing linking to libcap-dev and later I realised that the dependency is not properly set in the package.xml

https://github.com/ros-controls/control_toolbox/actions/runs/11671849281/job/32499176107

Sorry for the inconvenience


christophfroehlich (approved):
Thx for having a look

PR #190, opened 2024-11-06
Use pthread_setaffinity_np for setting affinity rather than sched_setaffinity

saikishor:
This PR aims to change the earlier approach with `pthread_setaffinity_np` as it is more generic and it can be scalable to all other controllers. 

The good thing is it doesn't affect the recently merged changes : https://github.com/ros-controls/ros2_control/pull/1852

I've also added corresponding tests in different places

### DIFF
--- a/src/realtime_helpers.cpp
+++ b/src/realtime_helpers.cpp
@@ -115,7 +115,7 @@ bool lock_memory(std::string & message)
 #endif
 }
 
-std::pair<bool, std::string> set_thread_affinity(int pid, int core)
+std::pair<bool, std::string> set_thread_affinity(pthread_t thread, int core)
 {
   std::string message;
 #ifdef _WIN32
@@ -158,8 +158,8 @@ std::pair<bool, std::string> set_thread_affinity(int pid, int core)
       CPU_SET(i, &cpuset);
     }
     // And actually tell the schedular to set the affinity of the thread of respective pid
-    const auto result =
-      set_affinity_result_message(sched_setaffinity(pid, sizeof(cpu_set_t), &cpuset), message);
+    const auto result = set_affinity_result_message(
+      pthread_setaffinity_np(thread, sizeof(cpu_set_t), &cpuset), message);
     return std::make_pair(result, message);
   }
 
@@ -167,8 +167,8 @@ std::pair<bool, std::string> set_thread_affinity(int pid, int core)
     // Set the passed core to the cpu set
     CPU_SET(core, &cpuset);
     // And actually tell the schedular to set the affinity of the thread of respective pid
-    const auto result =
-      set_affinity_result_message(sched_setaffinity(pid, sizeof(cpu_set_t), &cpuset), message);
+    const auto result = set_affinity_result_message(
+      pthread_setaffinity_np(thread, sizeof(cpu_set_t), &cpuset), message);
     return std::make_pair(result, message);
   }
   // Invalid core number passed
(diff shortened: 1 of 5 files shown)


saikishor:
@firesurfer can you check this once?

src/realtime_helpers.cpp
```diff
@@ -115,7 +115,7 @@ bool lock_memory(std::string & message)
 #endif
 }
 
-std::pair<bool, std::string> set_thread_affinity(int pid, int core)
+std::pair<bool, std::string> set_thread_affinity(pthread_t thread, int core)
```
firesurfer, line 118: In the header you used `std::thread`

firesurfer (approved):
In general this looks good to me with a minor notes:
https://en.cppreference.com/w/cpp/thread/thread/native_handle native_handle is only available for systems with a posix threading model. I am not sure if there might be some weird combination of systems (apart from windows) where it is not available and might lead to compilation errors

src/realtime_helpers.cpp
```diff
@@ -180,9 +180,18 @@ std::pair<bool, std::string> set_thread_affinity(int pid, int core)
 #endif
 }
 
+std::pair<bool, std::string> set_thread_affinity(std::thread & thread, int core)
+{
+  if (!thread.joinable()) {
```
firesurfer, line 185: Just out of interest: Why does a thread need to be joinable in order to set its affinity?

saikishor: Its just to kinda check if the thread is initialized or not. If not, there is no other way to check apart from that. 

If it is not joinable, then it's not active, so it doesn't make sense IMO

saikishor:
> In general this looks good to me with a minor notes: https://en.cppreference.com/w/cpp/thread/thread/native_handle native_handle is only available for systems with a posix threading model. I am not sure if there might be some weird combination of systems (apart from windows) where it is not available and might lead to compilation errors

Most of the systems now supports. The following information, I got it using Copilot

Systems that support a POSIX threading model (also known as POSIX threads or Pthreads) include most Unix-like operating systems. POSIX threads are a standard for thread creation and synchronization, defined by the POSIX.1c standard (IEEE Std 1003.1c-1995). Here are some systems that support POSIX threads:


**Unix-like Operating Systems**

- Linux: Most Linux distributions support POSIX threads natively. The GNU C Library (glibc) provides the implementation of POSIX threads on Linux.
- macOS: macOS (formerly OS X) supports POSIX threads. The implementation is provided by the system's libc.
- FreeBSD: FreeBSD supports POSIX threads through its native threading library.
- NetBSD: NetBSD also supports POSIX threads.
- OpenBSD: OpenBSD includes support for POSIX threads.
- Solaris: Solaris (and its derivatives like OpenSolaris and illumos) supports POSIX threads.

**Other Operating Systems**

- AIX: IBM's AIX operating system supports POSIX threads.
- HP-UX: Hewlett-Packard's HP-UX supports POSIX threads.
- QNX: QNX, a real-time operating system, supports POSIX threads.
- VxWorks: VxWorks, another real-time operating system, supports POSIX threads.
- Windows: While Windows does not natively support POSIX threads, there are libraries like Pthreads-w32 that provide a POSIX threads API on top of the native Windows threading model.

As this is the case, I think this should be good to go

fmauch (approved):
This looks good and should give more flexibility ;-)

christophfroehlich (approved):
LGTM

bmagyar (approved):

PR #193, opened 2024-11-12
Overloading the set_thread_affinity method for Windows compatibility

SENAI-GilmarCorreia:
Just me and my Windows setup again :)

### DIFF
--- a/include/realtime_tools/realtime_helpers.hpp
+++ b/include/realtime_tools/realtime_helpers.hpp
@@ -29,9 +29,9 @@
 #ifndef REALTIME_TOOLS__REALTIME_HELPERS_HPP_
 #define REALTIME_TOOLS__REALTIME_HELPERS_HPP_
 
+#include <pthread.h>
 #include <string>
 #include <thread>
-#include <pthread.h>
 #include <utility>
 
 namespace realtime_tools
--- a/include/realtime_tools/realtime_helpers.hpp
+++ b/include/realtime_tools/realtime_helpers.hpp
@@ -29,10 +29,16 @@
 #ifndef REALTIME_TOOLS__REALTIME_HELPERS_HPP_
 #define REALTIME_TOOLS__REALTIME_HELPERS_HPP_
 
-#include <pthread.h>
 #include <string>
 #include <thread>
 #include <utility>
+
+#ifdef _WIN32
+#include <windows.h>
+using NATIVE_THREAD_HANDLE = HANDLE;
+#else
+using NATIVE_THREAD_HANDLE = pthread_t;
+#endif
 
 namespace realtime_tools
 {
@@ -70,7 +76,7 @@ bool lock_memory(std::string & message);
  * \returns a pair of a boolean indicating whether the operation succeeded or not
  * and a message describing the result of the operation
 */
-std::pair<bool, std::string> set_thread_affinity(pthread_t thread, int core);
+std::pair<bool, std::string> set_thread_affinity(NATIVE_THREAD_HANDLE thread, int core);
 
 /**
  * Configure the caller thread affinity - Tell the scheduler to prefer a certain
--- a/include/realtime_tools/realtime_helpers.hpp
+++ b/include/realtime_tools/realtime_helpers.hpp
@@ -88,7 +88,7 @@ std::pair<bool, std::string> set_current_thread_affinity(int core);
  * \ref https://stackoverflow.com/a/150971
  * \returns The number of processors currently online (available)
 */
-int get_number_of_available_processors();
+int64_t get_number_of_available_processors();
 
 }  // namespace realtime_tools
 
--- a/include/realtime_tools/realtime_helpers.hpp
+++ b/include/realtime_tools/realtime_helpers.hpp
@@ -30,6 +30,7 @@
 #define REALTIME_TOOLS__REALTIME_HELPERS_HPP_
 
 #include <string>
+#include <thread>
 #include <utility>
 
 namespace realtime_tools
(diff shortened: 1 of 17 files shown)


include/realtime_tools/realtime_helpers.hpp
```diff
@@ -29,6 +29,7 @@
 #ifndef REALTIME_TOOLS__REALTIME_HELPERS_HPP_
 #define REALTIME_TOOLS__REALTIME_HELPERS_HPP_
 
+#include <pthread.h>
```
saikishor, line 32: @SENAI-GilmarCorreia Do you need `pthread.h` here?

SENAI-GilmarCorreia: To compile successfully, I really need this.

If it’s not included, I get the following error:

```
C:\path\to\workspace\src\realtime_tools\include\realtime_tools/realtime_helpers.hpp(72,50): error C2065: 'pthread_t': undeclared identifier [C:\path\to\workspace\build\realtime_tools\realtime_tools.vcxproj]
C:\path\to\workspace\src\realtime_tools\include\realtime_tools/realtime_helpers.hpp(72,60): error C2146: syntax error: missing ')' before identifier 'thread' [C:\path\to\workspace\build\realtime_tools\realtime_tools.vcxproj]
  Building Custom Rule C:/path/to/workspace/src/realtime_tools/CMakeLists.txt
  realtime_helpers.cpp
C:\path\to\workspace\src\realtime_tools\include\realtime_tools/realtime_helpers.hpp(72,50): error C2065: 'pthread_t': undeclared identifier [C:\path\to\workspace\build\realtime_tools\thread_priority.vcxproj]
C:\path\to\workspace\src\realtime_tools\include\realtime_tools/realtime_helpers.hpp(72,60): error C2146: syntax error: missing ')' before identifier 'thread' [C:\path\to\workspace\build\realtime_tools\thread_priority.vcxproj]
```

saikishor: So, you had to add it for pthread_t. I've a good solution to this then. I'll try to add it to your PR when I get some time 

So, we can avoid that header for windows

src/realtime_helpers.cpp
```diff
@@ -115,13 +115,16 @@ bool lock_memory(std::string & message)
 #endif
 }
 
-std::pair<bool, std::string> set_thread_affinity(pthread_t thread, int core)
-{
-  std::string message;
 #ifdef _WIN32
-  message = "Thread affinity is not supported on Windows.";
+std::pair<bool, std::string> set_thread_affinity(HANDLE thread, int core)
```
saikishor, line 119: @SENAI-GilmarCorreia Don't you need a definition of this method in the header file?

saikishor:
@SENAI-GilmarCorreia Do you mind if I push some changes to your branch?

SENAI-GilmarCorreia:
No problem! @saikishor

src/realtime_helpers.cpp
SENAI-GilmarCorreia: let me see If I can change this.

saikishor: No need. I'm already working on it now. I'll push it in few min

SENAI-GilmarCorreia:
@saikishor Done!

SENAI-GilmarCorreia:
@saikishor I've just fixed the issue by using the `__unix__` flag instead of `UNIX`, which wasn't defined correctly for non-Windows environments.

src/realtime_helpers.cpp
```diff
@@ -28,9 +28,7 @@
 
 #include "realtime_tools/realtime_helpers.hpp"
 
-#ifdef _WIN32
-#include <windows.h>
-#else
+#ifdef __unix__
```
aprotyas, line 31: I don't know which platforms we end up targeting or want to support, but [just leaving some text here](https://web.archive.org/web/20140625123925/http://nadeausoftware.com/articles/2012/01/c_c_tip_how_use_compiler_predefined_macros_detect_operating_system#:~:text=However%2C%20compilers%20for%20OSX%2C%20iOS%2C%20and%20Darwin%20%20do%20not%20define%20__unix__.%20To%20detect%20all%20BSD%20OSes%2C%20including%20OSX%2C%20iOS%2C%20and%20Darwin%2C%20use%20an%20#if/#endif%20that%20checks%20for%20__unix__%20along%20with%20__APPLE__%20and%20__MACH__%20(see%20the%20later%20section%20on%20OSX%20and%20iOS).) to ruminate on:

> However, compilers for OSX, iOS, and Darwin do not define `__unix__`. To detect all BSD OSes, including OSX, iOS, and Darwin, use an `#if`/`#endif` that checks for `__unix__` along with `__APPLE__` and `__MACH__` (see the later section on OSX and iOS).

saikishor:
```suggestion
#if defined(__unix__) || (defined(__APPLE__) && defined(__MACH__))
```
Make sense. I think we can use the following one as in the link you suggested  @aprotyas

saikishor:
@SENAI-GilmarCorreia I see we mention only `_WIN32`, should we also handle `_WIN64`?

SENAI-GilmarCorreia:
> @SENAI-GilmarCorreia I see we mention only `_WIN32`, should we also handle `_WIN64`?

Hello @saikishor. I have seen that _WIN32 supports both 32-bit and 64-bit Windows platforms in Microsoft Visual Studio.

saikishor (approved):

christophfroehlich (approved):

Issue #200, opened 2024-11-23
API of RealtimeBox is broken


christophfroehlich:
- ur_controller: https://github.com/UniversalRobots/Universal_Robots_ROS2_Driver/pull/1189
- ros2_controllers: https://github.com/ros-controls/ros2_controllers/actions/runs/11988629468/job/33424120547 fixed with https://github.com/ros-controls/ros2_controllers/pull/1385

the deprecation failed, things to fix

- [x] The `set`/`get` method is not compatible any more, @saikishor has an idea
- [x] `received_velocity_msg_ptr_.set(nullptr);` like we used in diff_drive_controller now builds but throws `C++ exception with description "bad_function_call" thrown in the test body.` at runtime
- [x] the realtime_box_best_effort.h is not available any more -> add it again pointing to the new one with a deprecation warning
- [x] add downstream builds to a CI job to be more aware of API breaking changes https://github.com/ros-controls/realtime_tools/pull/201

Not sure if we can fix that for the humble releases, or we have to revert that and branch for rolling.

PR #201, opened 2024-11-23
Add downstream build CI job

christophfroehlich:
#200


.github/workflows/binary-build.yml
```diff
@@ -26,3 +26,17 @@ jobs:
       ros_distro: ${{ matrix.ROS_DISTRO }}
       ros_repo: ${{ matrix.ROS_REPO }}
       ref_for_scheduled_build: master
+  build-downstream:
+    uses: ros-controls/ros2_control_ci/.github/workflows/reusable-industrial-ci-with-cache.yml@test_downstream
```
christophfroehlich, line 30:
```suggestion
    uses: ros-controls/ros2_control_ci/.github/workflows/reusable-industrial-ci-with-cache.yml@master
```

.github/workflows/binary-downstream-build.yml
```diff
@@ -0,0 +1,28 @@
...
+  pull_request:
+    branches:
+      - master
+  push:
+    branches:
+      - master
+
+jobs:
+  build-downstream:
+    uses: ros-controls/ros2_control_ci/.github/workflows/reusable-industrial-ci-with-cache.yml@test_downstream
```
christophfroehlich, line 16:
```suggestion
    uses: ros-controls/ros2_control_ci/.github/workflows/reusable-industrial-ci-with-cache.yml@master
```

christophfroehlich:
what is your opinion: Not testing and just building everything is fine here?

saikishor:
> what is your opinion: Not testing and just building everything is fine here?

I don't think we have to test anything for now. We just build and that's it and for the testing we have the nightly builds. If we have a use case in future, let's do that

saikishor (approved):
LGTM

fmauch:
> > what is your opinion: Not testing and just building everything is fine here?
> 
> I don't think we have to test anything for now. We just build and that's it and for the testing we have the nightly builds. If we have a use case in future, let's do that

I would agree in that case.

christophfroehlich:
@mergifyio backport humble

PR #204, opened 2024-11-24
Use windows CI build

christophfroehlich:
The main purpose is to avoid adding code which does not compile on Windows. Github provides windows-2019 runners, which can be used here. However, https://github.com/nektos/gh-act does not support this and I was only to test the workflow online.

- Binary releases are not published for rolling, so I hardcoded jazzy
- I added it now to be triggered at PRs and pushes, but it might be configured to run on comments/labels only.
- Needs https://github.com/ros-controls/ros2_control_ci/pull/175

### DIFF
--- a/.github/workflows/binary-build-win.yml
+++ b/.github/workflows/binary-build-win.yml
@@ -22,7 +22,7 @@ jobs:
     #   (github.event_name == 'issue_comment' && contains(github.event.comment.body, '/check-windows')) ||
     #   (github.event_name == 'pull_request' && contains(github.event.label.name, 'check-windows')) ||
     #   (github.event_name == 'workflow_dispatch')
-    uses: ros-controls/ros2_control_ci/.github/workflows/reusable-ros-tooling-win-build.yml@windows_ci
+    uses: ros-controls/ros2_control_ci/.github/workflows/reusable-ros-tooling-win-build.yml@master
     with:
       ros_distro: jazzy
       ref_for_scheduled_build: master
--- a/.github/workflows/binary-build-win.yml
+++ b/.github/workflows/binary-build-win.yml
@@ -1,0 +1,29 @@
+name: Binary Build Windows
+# author: Christoph Fröhlich <christoph.froehlich@ait.ac.at>
+# description: 'Build & test all dependencies from released (binary) windows packages.'
+
+on:
+  workflow_dispatch:
+  pull_request:
+    branches:
+      - master
+    # types:
+    #   - labeled
+  push:
+    branches:
+      - master
+  # issue_comment:
+  #   types:
+  #     - created
+
+jobs:
+  binary-windows:
+    # if: |
+    #   (github.event_name == 'issue_comment' && contains(github.event.comment.body, '/check-windows')) ||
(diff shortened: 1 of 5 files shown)


.github/workflows/binary-build-win.yml
```diff
@@ -0,0 +1,29 @@
...
+  #   types:
+  #     - created
+
+jobs:
+  binary-windows:
+    # if: |
+    #   (github.event_name == 'issue_comment' && contains(github.event.comment.body, '/check-windows')) ||
+    #   (github.event_name == 'pull_request' && contains(github.event.label.name, 'check-windows')) ||
+    #   (github.event_name == 'workflow_dispatch')
+    uses: ros-controls/ros2_control_ci/.github/workflows/reusable-ros-tooling-win-build.yml@windows_ci
```
christophfroehlich, line 25:
```suggestion
    uses: ros-controls/ros2_control_ci/.github/workflows/reusable-ros-tooling-win-build.yml@master
```

test/test_async_function_handler.hpp
```diff
@@ -24,9 +24,9 @@
 namespace realtime_tools
 {
 enum class return_type : std::uint8_t {
-  OK = 0,
-  ERROR = 1,
-  DEACTIVATE = 2,
```
saikishor, line 29: Can they be capital as usual?

test/test_async_function_handler.cpp
```diff
@@ -401,7 +401,7 @@ TEST_F(AsyncFunctionHandlerTest, check_exception_handling)
   async_class.get_handler().start_thread();
   trigger_status = async_class.trigger();
   ASSERT_TRUE(trigger_status.first);
-  ASSERT_EQ(realtime_tools::return_type::OK, trigger_status.second);
```
saikishor, line 404: same here and the rest of changes

test/test_async_function_handler.hpp
christophfroehlich: no, see the other comments

SENAI-GilmarCorreia:
Here's the corrected English version:

@christophfroehlich I am trying to understand the problem here. Do you want to revert the enums back to capital letters?

As I have seen, ERROR already exists in Windows systems because this identifier is reserved for error codes in Windows.

You can simply rename it by changing the case, as you did, or you can change the name of the enum:

```cpp
enum class return_type : std::uint8_t {
  OK = 0,
  FAILURE = 1,
  DEACTIVATE = 2,
};
```

I'm testing here as well.

christophfroehlich:
Thanks for reaching out. Exactly, if only the ERROR is a problem then I'd rename it as you suggested. Or maybe there is a compile definition to avoid this, like NOMINMAX ;)

SENAI-GilmarCorreia:
As I've seen, this definition is really important. You can use #undef ERROR, but it might cause some unexpected behaviors.

SENAI-GilmarCorreia (approved):

fmauch (approved):
This looks fine with me

christophfroehlich:
@mergifyio backport humble

PR #206, opened 2024-11-25
Move the header files to `.hpp` extension

saikishor:
As a standard now in ROS 2, all C++ headers need to be in .hpp format. So, I've moved some pending headers to .hpp extension


christophfroehlich (approved):
Great!

urfeex (approved):
Loving it!

saikishor:
> Would you mind creating good-first-issues in all our repositories to use the new headers?

Sure, I will do it during the day

PR #208, opened 2024-11-27
Add support to parse multiple cores for setting CPU affinity

saikishor:
This PR adds a new method that helps to set multiple CPUs for affinity instead of only one and this will help in choosing couple of CPUs to run the controller_manager etc


firesurfer:
I just saw the PR by chance and was interested as I authored the original one. I am wondering in which (realtime) scenario would you like to prefer a set of cores? From my understanding every time a thread is executed on the different core we suffer from

a) Context switch overhead
b) Additional cache misses

Both brings in an uncertainty which we do not want in RT scenarios. 


The only case I can imagine this could help is modern CPUs with multiple dies inside (e.g. most Ryzen processors). But the OS scheduler will already take care of trying not to move threads between cores on different dies per default.

include/realtime_tools/realtime_helpers.hpp
```diff
@@ -101,6 +115,17 @@ std::pair<bool, std::string> set_thread_affinity(std::thread & thread, int core)
 */
 std::pair<bool, std::string> set_current_thread_affinity(int core);
 
+/**
+ * Configure the current thread affinity - Tell the scheduler to prefer a certain
+ * core for the current thread.
+ * \note The threads created by the calling thread will inherit the affinity.
+ * \param[in] core the cpu numbers of the core. If an empty vector is passed,
```
firesurfer, line 122: I think you missed updating the comments

saikishor: Well isn't it how it works. Previously it is -1 and now if the vector is empty should behave same

saikishor:
@firesurfer You can set it at the startup as we do in the ros2_control_node or in the configure of controllers and hardware components

include/realtime_tools/realtime_helpers.hpp
firesurfer: Ah sorry I clicked in the wrong line:

> Tell the scheduler to prefer a certain
 * core for the current thread.
 
 This is what I meant. Should probably be: > Tell the scheduler to prefer a certain set of cores for the current thread

firesurfer:
> @firesurfer You can set it at the startup as we do in the ros2_control_node or in the configure of controllers and hardware components

I am aware of that. But I just wanted to understand why it is an advantage in a RT scenario to use the group of cores instead of tying the thread to a single core.

saikishor:
> I am aware of that. But I just wanted to understand why it is an advantage in a RT scenario to use the group of cores instead of tying the thread to a single core.

By default, it uses mostly Core 0 as other. If you can use a set of performance cores, it would be better for the RT application, if your application needs more computational power than one core, then this definitely helps

christophfroehlich (approved):
LGTM

christophfroehlich:
@mergifyio backport humble

PR #210, opened 2024-11-28
First step towards modernizing the rt publisher

firesurfer:
I think the rt publisher is one of the most commonly used tools in this package.
When ever I was using it I stumbled upon its "weird" API. 

This PR is a first step towards modernizing the RealtimePublisher class, providing a tiny bit of comfort by adding a `tryPublish` method and providing some typedefs similar to the rclcpp::Publisher.

I plan to add a follow up PR where I refactor the NON_POLLING define into a template parameter.


include/realtime_tools/realtime_publisher.hpp
```diff
@@ -112,6 +124,20 @@ class RealtimePublisher
...
+   * Tries to gain unique access to msg_ variable. If this succeeds
+   * update the msg_ variable and call unlockAndPublish
+   * @return false in case no lock for the realtime variable could be acquired
+   */
+  bool tryPublish(const MessageT & msg)
+  {
+    if (!trylock()) return false;
+
+    msg_ = msg;
+    unlockAndPublish();
```
aprotyas, line 138: This is missing a `return true;` I suppose?

include/realtime_tools/realtime_publisher.hpp
```diff
@@ -210,12 +236,12 @@ class RealtimePublisher
   std::condition_variable updated_cond_;
 #endif
 
-  enum { REALTIME, NON_REALTIME, LOOP_NOT_STARTED };
-  std::atomic<int> turn_;  // Who's turn is it to use msg_?
+  enum class State : int { REALTIME, NON_REALTIME, LOOP_NOT_STARTED };
```
aprotyas, line 239: You can use a smaller backing type here since there are only 3 cases.

```suggestion
  enum class State : uint8_t { REALTIME, NON_REALTIME, LOOP_NOT_STARTED };
```

include/realtime_tools/realtime_publisher.hpp
firesurfer: Good eyes! Thanks for pointing this out.

include/realtime_tools/realtime_publisher.hpp
firesurfer: On the other hand we also don't really profit from using a smaller datatype. In the end it will be filled up to an int on 90% of the systems.

firesurfer:
@christophfroehlich anything missing in order to get this merged?
The CI checks fail due to some flaky tests of the AsyncFunctionHandler.

christophfroehlich:
another review ;)

firesurfer:
And I just added a test for the tryPublish method :) 

@saikishor would you mind reviewing this PR ?

saikishor:
> And I just added a test for the tryPublish method :) 
> 
> @saikishor would you mind reviewing this PR ? 

I'll check it during the day. I have some pending review already

saikishor (approved):
In general LGTM!

I left some minor comments

saikishor (approved):

Issue #24, opened 2018-01-08
Latched RealtimePublisher is racy on startup


mikepurvis:
Discovered this in my application— didn't matter whether I made the initial `unlockAndPublish` call from the realtime or non-realtime thread, it still would only go through sometimes. Obviously once everything's properly initialized then the most recent message is properly persisted and it doesn't matter, but it would be nice to reliably see that initial latched message.

I think this could be handled in client code by checking `is_running()`, except that that method is private; the ideal solution would be for `unlockAndPublish` itself to make a check somehow, and if the publish loop is not up or not yet waiting on the condition, set a flag so that it knows there's an initial message to be published (or even just introduce a separate method which can be used to seed the publisher from non-realtime with an initial message).

PR #211, opened 2024-11-28
Refactor/rt pub define

firesurfer:
This is a followup PR for #210 

The approach in general probably needs some discussion. But I think it is a first step into removing conditional compilation based on a define set in a different file. 

This implementation keeps the ability to select the polling mode via the define but prints a deprecation message.

### DIFF
--- a/include/realtime_tools/realtime_publisher.hpp
+++ b/include/realtime_tools/realtime_publisher.hpp
@@ -49,27 +49,39 @@
 
 namespace realtime_tools
 {
-template <class Msg>
+template <class MessageT>
 class RealtimePublisher
 {
-private:
-  using PublisherSharedPtr = typename rclcpp::Publisher<Msg>::SharedPtr;
-
 public:
+  /// Provide various typedefs to resemble the rclcpp::Publisher type
+  using PublisherType = rclcpp::Publisher<MessageT>;
+  using PublisherSharedPtr = typename rclcpp::Publisher<MessageT>::SharedPtr;
+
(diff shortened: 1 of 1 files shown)


include/realtime_tools/realtime_publisher.hpp
```diff
@@ -49,173 +49,336 @@
...
+// Allow to select the way the RT Publisher is implemented
+enum class RealtimePublisherPollingMode { Polling, NonPolling };
+
+// Provide backwards compatibility for NON_POLLING define
+#ifndef NON_POLLING
+#define NON_POLLING RealtimePublisherPollingMode::Polling
+#else
+#undef NON_POLLING
+#define NON_POLLING RealtimePublisherPollingMode::NonPolling
+// Note this is the only way to somehow deprecate a define as #warning is not portable
```
christophfroehlich, line 62: Why not doing something like here?
https://github.com/ros-controls/realtime_tools/blob/3ee84c57c2530820224b25f7b28f50f3901f4bf8/include/realtime_tools/realtime_box.h#L20-L27

firesurfer: We can also go for that way. I assume it works for all supported platforms.

christophfroehlich:
@firesurfer do you have capacity right now to finish this PR? We have a compile-time warning from the tests since #210, would be great to fix at least this one ;) 
Have you by chance read #24 and #81? Not sure if the first one still applies, but the restructuring is still valid, right?

PR #217, opened 2024-12-04
Changes after branching humble

christophfroehlich:


.github/workflows/humble-abi-compatibility.yml
```diff
@@ -8,7 +8,7 @@ on:
 concurrency:
   # cancel previous runs of the same workflow, except for pushes on humble branch
   group: ${{ github.workflow }}-${{ github.ref }}
-  cancel-in-progress: ${{ github.ref != 'humble' }}
+  cancel-in-progress: ${{ !startsWith(github.ref, '/refs/heads') }}
```
Raivias, line 11: This lets any branch head continue. Is that what you intend?

It's probably fine since this repo seems to keep a small set of branches.

Raivias (approved):

.github/workflows/humble-abi-compatibility.yml
christophfroehlich: what I want here is that old workflows after pushes to PRs get cancelled, but wfs after pushes to the branches (master, humble..) aren't. so that we save CI time in PRs but have tests for every commit on the head branches.

saikishor (approved):
LGTM

christophfroehlich:
@mergifyio backport humble

PR #225, opened 2024-12-07
[Humble] Move the header files to `.hpp` extension (#206) - No deprecation notice

saikishor:


christophfroehlich (approved):
Do we annoy users if we add a message ([msvc](https://learn.microsoft.com/de-de/cpp/preprocessor/message?view=msvc-170), [gcc](https://gcc.gnu.org/onlinedocs/gcc/Diagnostic-Pragmas.html) but not sure for clang) instead of warnings like we do on newer distros? At least compilation would not break with -Werr then.

saikishor:
> Do we annoy users if we add a message ([msvc](https://learn.microsoft.com/de-de/cpp/preprocessor/message?view=msvc-170), [gcc](https://gcc.gnu.org/onlinedocs/gcc/Diagnostic-Pragmas.html) but not sure for clang) instead of warnings like we do on newer distros? At least compilation would not break with -Werr then.

Sure! I can add a message instead of a warning

saikishor:
> Do we annoy users if we add a message ([msvc](https://learn.microsoft.com/de-de/cpp/preprocessor/message?view=msvc-170), [gcc](https://gcc.gnu.org/onlinedocs/gcc/Diagnostic-Pragmas.html) but not sure for clang) instead of warnings like we do on newer distros? At least compilation would not break with -Werr then.

```
In file included from /home/user/rolling_worspace/ros2_control_ws/src/ros2_control/controller_manager/test/test_chainable_controller/test_chainable_controller.hpp:25,
                 from /home/user/rolling_worspace/ros2_control_ws/src/ros2_control/controller_manager/test/test_controller_manager_srvs.cpp:30:
/home/user/rolling_worspace/ros2_control_ws/src/realtime_tools/include/realtime_tools/realtime_buffer.h:22:100: note: ‘#pragma message: This header include is deprecated. Please update your code to use 'realtime_buffer.hpp' header.’
   22 |   "This header include is deprecated. Please update your code to use 'realtime_buffer.hpp' header.")  //NOLINT

PR #228, opened 2024-12-09
Allow to get thread handle in `RealtimePublisher`

### LINKED ISSUE #176: Allow to rename thread in `RealtimePublisher`
At least in Linux, it is possible to rename a thread, via calls like [ptrhead_setname_np](https://man7.org/linux/man-pages/man3/pthread_setname_np.3.html).

For application featuring many `RealtimePublishers` and many other threads, diving into `perf record` outputs (e.g. via `hotspot` tool) is difficult because by default all threads have the same name (First 15 characters of the process name), and thread renaming is a powerful tool to discriminate at a glance which thread does what.

To allow this, either the implementation of `RealtimePublisher` shall permit to obtain the native handle (see [this](https://en.cppreference.com/w/cpp/thread/thread/native_handle)) or may enable for Linux only an additional method that uses `pthread_setname_np`.

I can submit a PR depending on maintainers feedback of course 👍🏻

roncapat:
Trying to close #176 as suggested by @firesurfer.


include/realtime_tools/realtime_publisher.hpp
```diff
@@ -202,8 +205,6 @@ class RealtimePublisher
   std::atomic<bool> is_running_;
   std::atomic<bool> keep_running_;
 
-  std::thread thread_;
-
   std::mutex msg_mutex_;  // Protects msg_
```
firesurfer, line 208: I would suggest to also expose (at least) the mutex if we expose the thread.

include/realtime_tools/realtime_publisher.hpp
```diff
@@ -153,6 +153,9 @@ class RealtimePublisher
 #endif
   }
 
+protected:
+  std::thread thread_;
+
```
saikishor, line 158:
```suggestion
  std::thread & get_thread() { return thread_; }
  
  const std::thread & get_thread() const { return thread_; }

```

saikishor (changes requested):
Why not something like this?

include/realtime_tools/realtime_publisher.hpp
saikishor: This way you don't have to make it protected and without inheritance you might be able to rename it

roncapat: Ah yes, it would work for me :)

include/realtime_tools/realtime_publisher.hpp
```diff
@@ -202,8 +206,6 @@ class RealtimePublisher
   std::atomic<bool> is_running_;
   std::atomic<bool> keep_running_;
 
-  std::thread thread_;
-
```
roncapat, line 206: Need to undo these two removed lines tomorrow from PC

include/realtime_tools/realtime_publisher.hpp
firesurfer: This was actually my suggestion in #176. 
I think it is not a good idea to expose the thread publicly as it would allow to interact with the class internals
E.g. joining, detaching

saikishor: I understand your point, but inheriting to just have this renaming and other stuff would be cumbersome right?

Moreover, if the user does it with that method, then he is doing something wrong which is not supposed to be. May be we can document it to avoid such things. 

What do you think? @firesurfer?

include/realtime_tools/realtime_publisher.hpp
```diff
@@ -202,8 +206,6 @@ class RealtimePublisher
   std::atomic<bool> is_running_;
   std::atomic<bool> keep_running_;
```
roncapat, line 208:
```suggestion
  std::thread thread_;

```

include/realtime_tools/realtime_publisher.hpp
```diff
@@ -201,7 +205,6 @@ class RealtimePublisher
   PublisherSharedPtr publisher_;
   std::atomic<bool> is_running_;
   std::atomic<bool> keep_running_;
```
roncapat, line 207:
```suggestion
  std::atomic<bool> keep_running_;

```

saikishor:
@roncapat Can you please fix the pre-commit?

saikishor (approved):
LGTM

PR #230, opened 2024-12-10
Remove duplicate wf

christophfroehlich:
Mistake with #217


saikishor (approved):
Makes sense. 
Thanks for fixing it

Issue #198, opened 2024-11-23
Flaky tests of AsyncFunctionHandlerTest


christophfroehlich:
```
    [ RUN      ] AsyncFunctionHandlerTest.check_triggering
    [WARN] [1732329912.212673771] [AsyncFunctionHandler]: Could not enable FIFO RT scheduling policy. Consider setting up your user to do FIFO RT scheduling. See [https://control.ros.org/master/doc/ros2_control/controller_manager/doc/userdoc.html] for details.
    /home/runner/work/realtime_tools/realtime_tools/.work/target_ws/src/realtime_tools/test/test_async_function_handler.cpp:140: Failure
    Value of: trigger_status.first
      Actual: false
    Expected: true
    
    [  FAILED  ] AsyncFunctionHandlerTest.check_triggering (0 ms)
```
https://github.com/ros-controls/realtime_tools/actions/runs/11983194146/job/33412345900
https://github.com/ros-controls/ros2_control_ci/actions/runs/11983202628/job/33412367563

christophfroehlich:
```
     [ RUN      ] AsyncFunctionHandlerTest.trigger_for_several_cycles
    [WARN] [1729911511.729861234] [AsyncFunctionHandler]: Could not enable FIFO RT scheduling policy. Consider setting up your user to do FIFO RT scheduling. See [https://control.ros.org/master/doc/ros2_control/controller_manager/doc/userdoc.html] for details.
    /home/runner/work/realtime_tools/realtime_tools/.work/target_ws/src/realtime_tools/test/test_async_function_handler.cpp:176: Failure
    Expected: (missed_triggers) < (static_cast<int>(0.001 * total_cycles)), actual: 337 vs 100
    The missed triggers cannot be more than 0.1%!
    
    [  FAILED  ] AsyncFunctionHandlerTest.trigger_for_several_cycles (11867 ms)
```
https://github.com/ros-controls/realtime_tools/actions/runs/11528036924

PR #235, opened 2024-12-13
sleep after starting thread to fix flaky tests

saikishor:
Fixes https://github.com/ros-controls/realtime_tools/issues/198
I've tested it running it with arg `--retest-until-fail 200`


christophfroehlich (approved):
Thanks, let's give it a try :)

Issue #14, opened 2015-02-04
Lock-free buffer implementation


mathias-luedtke:
The current buffer implementation is not lock-free and does not support writes from the real-time part.
I have  written a proof-of-concept implementation that might be a viable replacement.
It enforces a last-write-wins strategy, since this is what we need for the controlles, right?

You can find it here: https://gist.github.com/mathias-luedtke/773f043a31b3bcce3598
It is based on boost::atomic and boost::lockfree and requires a recent boost installation (14.04/indigo works).
[rosrt](http://wiki.ros.org/rosrt) backported both to hydro, but is not maintained anymore.

It is divided into two parts: _LockfreeBuffer_ and _LockedBuffer_.

_LockfreeBuffer_ maintains a pre-allocated lock-free freelist and provides the most recent data (push) to one single reader (pop).
- all calls should be lock-free, thread-safe and realtime-safe (if the payload copy constructor is!)
- Concurrent reads will lead to an empty pointer for one reader.
- The data is provided with RAII pointers to prevent misuses.
- The freelist must be big enough (number of writers + 1 reader) to work without starvation.
- Data can be pushed and pop from real-time and non-real-time parts.

However, the LockfreeBuffer only works for single consumers (sufficient for all current ros_controllers).
Therefore I have provided a _LockedBuffer_ implementation, that runs on top of the LockfreeBuffer.
- blocking read (~ ReatimeBuffer::readFromNonRT) 
- non-blocking try_read (~ ReatimeBuffer::readFromRT).
- write is just an alias for LockfreeBuffer::push

It is not ready for release yet, but I would really appreciate your comments on this.

adolfo-rt:
Allow me a day or two to read the code in detail. From the summary, I'm concerned about the multiple reader case, if reading is subject to blocking, then it largely defeats the purpose of the data structure, doesn't it?. At any rate, I haven't  yet dug into the code.

I would suggest looking into at least two high-quality implementations of lock-free data structures I know of:

#### Orocos RTT

They implement multi-reader, multi-writer synchronization primitives with different locking and buffering policies. Take a look at [these classes](http://www.orocos.org/stable/documentation/rtt/v2.x/api/html/classRTT_1_1base_1_1DataObjectInterface.html), which implement the core data structures, in particular [RTT::base::DataObjectLockFree](
http://www.orocos.org/stable/documentation/rtt/v2.x/api/html/classRTT_1_1base_1_1DataObjectLockFree.html).

[...]

PR #253, opened 2025-01-01
Add Lock-free Queue

saikishor:
Fixes https://github.com/ros-controls/realtime_tools/issues/14


test/lock_free_queue_tests.cpp
```diff
@@ -0,0 +1,487 @@
...
+  producer.join();
+  done = true;
+  consumer.join();
+
+  ASSERT_EQ(producer_count, consumer_count);
+  ASSERT_EQ(producer_count, iterations);
+  ASSERT_EQ(consumer_count, iterations);
+}
+
+// TEST(LockFreeSPSCQueue, test_lockfree_queue_bounded_push)
```
christophfroehlich, line 222: ?

christophfroehlich (commented):
TBH, I don't fully understand all the C++ specifics here :grimacing: 
But the API seems to be clean from a user perspective, looking at the tests.

- Does this replace the RealtimeBuffer? Or should the RealtimeBuffer be rewritten to use this queue with capacity of 1?
- What is the practical difference between the RealtimeBox and the queue with capacity of 1? Maybe this is clear for some, but I'd need (and appreciate) a user guide when to use what.
- If we deprecate something, we can also add migration notes to this repository and link it on control.ros.org

package.xml
```diff
@@ -19,6 +19,7 @@
 
   <buildtool_export_depend>ament_cmake</buildtool_export_depend>
 
+  <build_depend>libboost-dev</build_depend>
```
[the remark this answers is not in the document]
saikishor: You are right, it should also be `build_export_depend` and `exec_depend`

test/lock_free_queue_tests.cpp
saikishor: This is the test that needs to be removed for multithreaded bounded queue for SPSC Queue. I'll do the cleanup now

saikishor: Done

saikishor:
@christophfroehlich First of all, thanks for reviewing it

> * Does this replace the RealtimeBuffer? Or should the RealtimeBuffer be rewritten to use this queue with capacity of 1?

I believe yes, we can replace this with the RealtimeBuffer. Ideally, it should serve as a replacement. I wouldn't mix them to add backward compatibility. I think it is better to separate both approaches.

> * What is the practical difference between the RealtimeBox and the queue with capacity of 1? Maybe this is clear for some, but I'd need (and appreciate) a user guide when to use what.

IMO, RealtimeBox is something you have to use, if you want to protect the data from concurrent access. For me, it is something to having the same thing protected by the mutex. However, to be frank I'm very skeptical of it being used with the shared_ptrs, because you can still access the contents outside of the RealtimeBox and this is couldn't be safe at all. 

Queue of capacity one might not be a good thing to use here, as it is meant to be used for buffer operations. Moreover, a queue of 1 is not practical, as soon as you have an element, you cannot push elements unless you have a consumer. Ofcourse, this can be done with bounded_push.

> * If we deprecate something, we can also add migration notes to this repository and link it on control.ros.org

If it's okay for you. I would separate deprecation into a different PR as this is not the main intention of this PR. Once we have it cleanly integrated with ros2_controllers, then we can cleanly deprecate it. What do you think?

test/lock_free_queue_tests.cpp
```diff
@@ -0,0 +1,441 @@
...
+    }
+  });
+
+  std::this_thread::sleep_for(std::chrono::milliseconds(100));
+
+  std::cerr << "producer_count: " << producer_count << std::endl;
+
+  std::thread consumer([&]() {
+    int value;
+    while (!done) {
```
erickisos, line 420: Can we merge those two conditions?
```suggestion
    while (!done && queue.pop(value)) {
```

include/realtime_tools/lock_free_queue.hpp
```diff
@@ -0,0 +1,358 @@
...
+   * the operating system.
+   * @link https://www.boost.org/doc/libs/1_74_0/doc/html/boost/lockfree/queue.html
+   */
+  bool is_lock_free() const
+  {
+    if constexpr (is_spsc_queue<LockFreeContainer>::value) {
+      return true;
+    } else {
+      return data_queue_.is_lock_free();
+    }
```
erickisos, line 312:
```suggestion
    return constexpr (is_spsc_queue<LockFreeContainer>::value) || data_queue_.is_lock_free();
```

include/realtime_tools/lock_free_queue.hpp
```diff
@@ -0,0 +1,358 @@
...
+   * @return false If the data could not be pushed
+   * @note This function is enabled only if the queue is of multiple producer and multiple consumer
+   * type and only if the data type is convertible to the template type of the queue
+   * @note Can be used in a multi threaded applications
+   */
+  template <
+    typename U, bool IsSPSCQueue = is_spsc_queue<LockFreeContainer>::value,
+    typename std::enable_if_t<!IsSPSCQueue, int> = 1>
+  [[nodiscard]] std::enable_if_t<std::is_convertible_v<T, U>, bool> bounded_push(const U & data)
+  {
```
erickisos, line 236: I might be missing something, but I see this as kinda the same function in [line 203](https://github.com/ros-controls/realtime_tools/pull/253/files#diff-ea7cd06f818c3d7ad61d89fe4cee66ec3c3734c83050862a7754d01768e4e6a4R203), is there any reason to do this separately?

maybe it would be better if we handle the retries with a while loop and a predefined (configurable) limit

include/realtime_tools/lock_free_queue.hpp
saikishor: I've removed the `constexpr` in your suggestions

include/realtime_tools/lock_free_queue.hpp
saikishor: Done

saikishor: Thanks for the suggestions. I applied to other functions as well

christophfroehlich (dismissed):
I tested that with https://github.com/ros-controls/ros2_controllers/pull/1480, the API works for me!

saikishor:
> I tested that with [ros-controls/ros2_controllers#1480](https://github.com/ros-controls/ros2_controllers/pull/1480), the API works for me!

@christophfroehlich I had to rewrite some parts of tests to reuse rather than copying for every instance. I've add more tests for all constructable types

package.xml
```diff
@@ -19,6 +19,7 @@
 
   <buildtool_export_depend>ament_cmake</buildtool_export_depend>
 
+  <depend>libboost-dev</depend>
```
bmagyar, line 22: is there a smaller dependency by any chance?

saikishor: https://github.com/ros/rosdistro/blob/master/rosdep/base.yaml#L2833-L2842

It is basic core one

If [boost](https://github.com/ros/rosdistro/blob/master/rosdep/base.yaml#L376-L392) is included, then it pulls in everything

bmagyar (approved):

PR #261, opened 2025-01-20
Remove RealtimeClock

### LINKED ISSUE #23: Hard coded loop rate
Forgive my ignorance if my issue is interpreting the code incorrectly as i'm pretty new to realtime_tools. In [this line](https://github.com/ros-controls/realtime_tools/blob/a5467ed2371abaab7fa646a51a9db72bc8b4771f/src/realtime_clock.cpp#L103) the loop rate is hard coded to 750Hz. To me this means that if you're running at loop rates greater than 750Hz then you should probably not use this clock tool correct? It seems it would be very easy to make this a configurable argument perhaps with a default rate of 750Hz to not break backwards compatibility. 

Additionally i'm a bit confused about the realtime clock implementation here. I'm unable to find anywhere in any documentation here or otherwise that ```ros::Time::now()``` is not a real time safe call to make. If it's not safe, then I do see the use of this except the return of ```getSystemTime``` is a ```ros::Time``` object that, to get time I would still need to call ```::now()```. 

Sorry for the questions, but I'm working on a hard real time system so i'm very interested in making sure my method of getting time is a real time safe call. Thanks!

### LINKED ISSUE #241: Is RealtimeClock() still needed with ROS 2?
We don't use it anywhere.

Reading the discussion from #23, are those now realtime safe?

- `std::chrono::steady_clock::now()`
- `rclcpp::Clock().now()`
- `rclcpp::Clock::SharedPtr rclcpp::node_interfaces::NodeClockInterface::get_clock()`-> now()

christophfroehlich:
RealtimeClock got deprecated on Jazzy and Rolling with their sync within the last days.

Fixes #241 and closes #23


saikishor (approved):
As we already did a release deprecating it.
I think we should be fine as we discussed

saikishor (approved):

PR #262, opened 2025-01-20
Remove deprecated code

christophfroehlich:
Deprecations are released on Jazzy and Rolling with their sync within the last days.

Fixes #203

PR #263, opened 2025-01-29
Branch for jazzy

christophfroehlich:
we don't want to remove the deprecated code from jazzy #261 #262


.github/mergify.yml
```diff
@@ -8,6 +8,15 @@ pull_request_rules:
...
           - humble
 
+  - name: Backport to jazzy at reviewers discretion
+    conditions:
+      - base=master
+      - "label=backport-jazzy"
+    actions:
+      backport:
+        branches:
+          - humble
```
saikishor, line 18:
```suggestion
          - jazzy
```

saikishor (changes requested):

.github/mergify.yml
saikishor: @christophfroehlich rest looks good to me

christophfroehlich: oops, thanks!

saikishor (approved):
LGTM

Issue #212, opened 2024-11-28
Realtime Publisher - Select one behavior (Polling or NonPolling) and remove the other one


firesurfer:
I just refactored the RealtimePublisher in #211 and I am wondering if we really need these two different behaviors of the RT publisher. 

They make the code harder to read and maintain. In accordance with #199 I would suggest to use the non polling behavior and remove the other one. 

Perhaps someone could point out a use case where one or the other is preferable?

firesurfer:
Only place in ros-controls the `NON_POLLING` define is set is in the tests of this repository:

https://github.com/search?q=org%3Aros-controls%20NON_POLLING&type=code

If I search more general I can only find it in this repo and forks of it 
https://github.com/search?q=+NON_POLLING+AND+realtime_tools&type=code

bmagyar:
Sorry for being late on this, @christophfroehlich has brought this to the Working Group meeting. 

I also agree that we should strive for using non-polling everywhere, especially now that it can be implemented in an elegant and transparent way.

PR #280, opened 2025-02-18
[RTPublisher] use NON_POLLING as default for the realtime pubisher 

saikishor:
Fixes: https://github.com/ros-controls/realtime_tools/issues/212

As reported by @bijoua29 

> I ran some tests with non-polling in some controllers and perf doesn't even report the sleep anymore. Most of the work is now in the DDS publish() method.
>
> I did another test where I compiled realtime_tools from source and set it to non-polling. When I ran my full stack, the CPU usage of ros2 control node reduced from ~75% to ~52% CPU i.e. around a 30% reduction in CPU usage. Note: this is percent of 1 CPU so since I have 32 CPUs the actual usage is 1/32 of that.


christophfroehlich (approved):
Thanks for the cleanup and even improving it. LGTM, but let's wait for some real-world tests

include/realtime_tools/realtime_publisher.hpp
```diff
@@ -196,37 +177,23 @@ class RealtimePublisher
...
-#ifdef NON_POLLING
-        updated_cond_.wait(lock_);
-#else
-        unlock();
-        std::this_thread::sleep_for(std::chrono::microseconds(500));
-        lock();
-#endif
+      {
+        // Locks msg_ and copies it to outgoing
+        std::unique_lock<std::mutex> lock_(msg_mutex_);
```
bijoua29, line 187: Not a big deal, but now that we're not locking and unlocking anymore, const std::lock_guard should be sufficient instead of std::unique_lock

bijoua29 (approved):

include/realtime_tools/realtime_publisher.hpp
bijoua29: Also, I ran a test pre- and post-PR, and I replicated the results from before on my software stack. That is, pre-PR, the ros2_control node was running at ~74% CPU usage and post-PR, it was running at 50% CPU usage.
Thanks again for the PR.

saikishor: unique_lock is being used for the conditional variable wait method

bijoua29: Ah, sure. Missed that. Ignore my comment.

destogl (approved):
Great work! Thanks!

PR #308, opened 2025-04-05
Fix startup race condition in realtime publisher

bijoua29:
I discovered a race condition when publishing a reliable message using a realtime publisher at startup. A typical pattern for a reliable topic is to create a publisher and then immediately publish the initial value like the following:

```
s_publisher_ = get_node()->create_publisher<TestMsg>("~/topic", transient_local_qos);
publisher_ = std::make_unique<RealtimePublisher<TestMsg>>(s_publisher_);

publisher_->lock();
// initialize message to publish
publisher_->unlockAndPublish();
```

unlockAndPublish() sets the state to realtime to notify the publishing loop before unlocking so it gets published in the publishing loop. The publisher loop thread first sets the state to nonrealtime and waits for the notification. The race condition is that the setting of state in unlockAndPublish() happens before the publishing loop initializes state so the notification is lost. In this case, the message doesn't get published. In my application, this is random and happens about half the time.

This PR fixes this race condition by inverting the order of initialization in the publisher loop. I also changed the access of is_running() to public. Then the following sequence always works:

```
s_publisher_ = get_node()->create_publisher<TestMsg>("~/topic", transient_local_qos);
publisher_ = std::make_unique<RealtimePublisher<TestMsg>>(s_publisher_);

while (!publisher_->is_running()) {
    std::this_thread::sleep_for(std::chrono::milliseconds(10));
}

publisher_->lock();
// initialize message to publish
publisher_->unlockAndPublish();
```

### DIFF
--- a/realtime_tools/include/realtime_tools/realtime_publisher.hpp
+++ b/realtime_tools/include/realtime_tools/realtime_publisher.hpp
@@ -185,12 +185,12 @@ public:
 
   const std::thread & get_thread() const { return thread_; }
 
+  bool is_running() const { return is_running_; }
+
 private:
   // non-copyable
   RealtimePublisher(const RealtimePublisher &) = delete;
   RealtimePublisher & operator=(const RealtimePublisher &) = delete;
-
-  bool is_running() const { return is_running_; }
 
   /**
    * \brief Publishing loop (runs in separate thread)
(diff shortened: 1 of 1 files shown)


christophfroehlich:
may this be the same as in #24?

saikishor:
I think what you have can happen. I'm not sure if this is the right fix🤔

christophfroehlich (commented):
Voting for #309

PR #309, opened 2025-04-05
Fix realtime publisher race condition upon initialization

saikishor:
Should fix : #24 
Closes: #308 

@bijoua29 can you test this and see if this helps?

### DIFF
--- a/realtime_tools/include/realtime_tools/realtime_publisher.hpp
+++ b/realtime_tools/include/realtime_tools/realtime_publisher.hpp
@@ -83,7 +83,7 @@ public:
     // This is important to ensure that the thread is properly initialized and ready to handle
     // messages before any other operations are performed on the RealtimePublisher instance.
     while (!thread_.joinable() ||
-           turn_.load(std::memory_order_acquire) != State::LOOP_NOT_STARTED) {
+           turn_.load(std::memory_order_acquire) == State::LOOP_NOT_STARTED) {
       std::this_thread::sleep_for(std::chrono::microseconds(100));
     }
   }
--- a/realtime_tools/include/realtime_tools/realtime_publisher.hpp
+++ b/realtime_tools/include/realtime_tools/realtime_publisher.hpp
@@ -116,14 +116,14 @@ public:
   /**
   * \brief Try to acquire the data lock for non-realtime message publishing
   *
-  * It first checks if the current state allows non-realtime message publishing (turn_ == NON_REALTIME)
+  * It first checks if the current state allows non-realtime message publishing (turn_ == REALTIME)
   * and then attempts to lock
   *
   * \return true if the lock was successfully acquired, false otherwise
   */
   bool trylock()
   {
-    if (turn_.load(std::memory_order_acquire) == State::NON_REALTIME && msg_mutex_.try_lock()) {
+    if (turn_.load(std::memory_order_acquire) == State::REALTIME && msg_mutex_.try_lock()) {
       return true;
     } else {
       return false;
@@ -159,7 +159,7 @@ public:
    */
   void unlockAndPublish()
   {
-    turn_.store(State::REALTIME, std::memory_order_release);
+    turn_.store(State::NON_REALTIME, std::memory_order_release);
     unlock();
   }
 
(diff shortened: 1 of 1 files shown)


realtime_tools/include/realtime_tools/realtime_publisher.hpp
```diff
@@ -78,6 +78,14 @@ class RealtimePublisher
   : publisher_(publisher), is_running_(false), keep_running_(true), turn_(State::LOOP_NOT_STARTED)
   {
     thread_ = std::thread(&RealtimePublisher::publishingLoop, this);
+
+    // Wait for the thread to be ready before proceeding
+    // This is important to ensure that the thread is properly initialized and ready to handle
+    // messages before any other operations are performed on the RealtimePublisher instance.
+    while (!thread_.joinable() ||
+           turn_.load(std::memory_order_acquire) != State::LOOP_NOT_STARTED) {
```
bijoua29, line 86: Maybe I don't understand this but shouldn't it be
turn_.load(std::memory_order_acquire) == State::LOOP_NOT_STARTED

realtime_tools/include/realtime_tools/realtime_publisher.hpp
```diff
@@ -116,14 +124,14 @@ class RealtimePublisher
...
-  * It first checks if the current state allows non-realtime message publishing (turn_ == NON_REALTIME)
+  * It first checks if the current state allows non-realtime message publishing (turn_ == REALTIME)
   * and then attempts to lock
   *
   * \return true if the lock was successfully acquired, false otherwise
   */
   bool trylock()
   {
-    if (turn_.load(std::memory_order_acquire) == State::NON_REALTIME && msg_mutex_.try_lock()) {
+    if (turn_.load(std::memory_order_acquire) == State::REALTIME && msg_mutex_.try_lock()) {
```
bijoua29, line 134: I feel like turn_ should be checked under mutex protection. So I would think the order here should be reversed. Although if the turn check fails you will need to release lock

realtime_tools/include/realtime_tools/realtime_publisher.hpp
```diff
@@ -196,33 +204,32 @@ class RealtimePublisher
...
   void publishingLoop()
   {
     is_running_ = true;
-    turn_.store(State::NON_REALTIME, std::memory_order_release);
 
     while (keep_running_) {
       MessageT outgoing;
 
       {
+        turn_.store(State::REALTIME, std::memory_order_release);
```
bijoua29, line 222: Similar to my comment above, I feel like turn_ should be checked when locked so this line should move under the unique_lock below. Again I could be wrong about this.

bijoua29 (commented):
Note my comments and see if you agree

bijoua29:
@saikishor Yes, I will test it today or tomorrow. Thanks for fixing this.

realtime_tools/include/realtime_tools/realtime_publisher.hpp
saikishor: You are right

saikishor, line 86:
```suggestion
           turn_.load(std::memory_order_acquire) == State::LOOP_NOT_STARTED) {
```

realtime_tools/include/realtime_tools/realtime_publisher.hpp
bijoua29: Oh yes, I forgot. I think you are right that it shouldn't be an issue.

christophfroehlich (approved):
This looks great.

So we don't need something like
```
while (!publisher_->is_running()) {
    std::this_thread::sleep_for(std::chrono::milliseconds(10));
}
```
in the application code, right?

saikishor:
> This looks great.
> 
> So we don't need something like
> ```
> while (!publisher_->is_running()) {
>     std::this_thread::sleep_for(std::chrono::milliseconds(10));
> }
> ```
> in the application code, right?

Yess! Exactly

bijoua29:
@saikishor I ran a test 20 times pre-PR and had a 85% failure rate. I ran another test 20 times post-PR and had 100% success. So I think the PR fixes the problem. I am approving the PR.

bijoua29 (approved):
LGTM

Issue #307, opened 2025-04-03
CI test hangs at ~RealtimePublisher()


Juliaj:
**Describe the bug**
I was able to reproduce the timeout reported in https://github.com/ros-controls/ros2_controllers/issues/1574 locally. The issue is very intermittent and generally not reproducible on demand. 

**To Reproduce**
1. run the unit tests for `force_torque_sensor_broadcaster`. 
2. Test hangs during `ForceTorqueSensorBroadcasterTest.InterfaceNames_Configure_Success`.

**Expected behavior**
Test doesn't hang.

**Environment (please complete the following information):**
 - OS: Ubuntu 
 - Jazzy

**Additional context**
gdb shows following stack when the test hung.

```
(gdb) bt
#0  0x000076fd5c4ecadf in __GI___clock_nanosleep (clock_id=clock_id@entry=0, flags=flags@entry=0, 
    req=0x7ffdbcceb530, rem=0x7ffdbcceb530) at ../sysdeps/unix/sysv/linux/clock_nanosleep.c:78
#1  0x000076fd5c4f9a27 in __GI___nanosleep (req=<optimized out>, rem=<optimized out>)
    at ../sysdeps/unix/sysv/linux/nanosleep.c:25
#2  0x00005ea440439ff7 in void std::this_thread::sleep_for<long, std::ratio<1l, 1000000l> >(std::chrono::duration<long, std::ratio<1l, 1000000l> > const&) ()
#3  0x00005ea440447a93 in realtime_tools::RealtimePublisher<geometry_msgs::msg::WrenchStamped_<std::allocator<void> > >::~RealtimePublisher() ()
#4  0x00005ea4404434b8 in std::default_delete<realtime_tools::RealtimePublisher<geometry_msgs::msg::WrenchStamped_<std::allocator<void> > > >::operator()(realtime_tools::RealtimePublisher<geometry_msgs::msg::WrenchStamped_<std::allocator<void> > >*) const ()
#5  0x00005ea44043f400 in std::unique_ptr<realtime_tools::RealtimePublisher<geometry_msgs::msg::WrenchStamped_<std::allocator<void> > >, std::default_delete<realtime_tools::RealtimePublisher<geometry_msgs::msg::WrenchStamped_<std::allocator<void> > > > >::~unique_ptr() ()
#6  0x00005ea4404393b4 in force_torque_sensor_broadcaster::ForceTorqueSensorBroadcaster::~ForceTorqueSensorBroadcaster() ()
[...]

PR #320, opened 2025-05-01
Fix the deadlock in the destructor of RealtimePublisher

saikishor:
Fixes: https://github.com/ros-controls/ros2_controllers/issues/1574
Fixes: https://github.com/ros-controls/ros2_controllers/issues/1669

Apart from this, I've seen a possible vulnerability with the current API, I'll open a different PR  to handle this one.

### DIFF
--- a/realtime_tools/include/realtime_tools/realtime_publisher.hpp
+++ b/realtime_tools/include/realtime_tools/realtime_publisher.hpp
@@ -99,13 +99,12 @@ public:
   /// Destructor
   ~RealtimePublisher()
   {
-    RCLCPP_DEBUG_STREAM(
-      rclcpp::get_logger("realtime_tools"), "Waiting for publishing thread to stop....");
+    RCLCPP_DEBUG(rclcpp::get_logger("realtime_tools"), "Waiting for publishing thread to stop....");
     stop();
     while (is_running()) {
       std::this_thread::sleep_for(std::chrono::microseconds(100));
     }
-    RCLCPP_DEBUG_STREAM(
+    RCLCPP_DEBUG(
       rclcpp::get_logger("realtime_tools"), "Publishing thread stopped, joining thread....");
     if (thread_.joinable()) {
       thread_.join();
--- a/realtime_tools/include/realtime_tools/realtime_publisher.hpp
+++ b/realtime_tools/include/realtime_tools/realtime_publisher.hpp
@@ -99,10 +99,14 @@ public:
   /// Destructor
   ~RealtimePublisher()
   {
+    RCLCPP_DEBUG_STREAM(
+      rclcpp::get_logger("realtime_tools"), "Waiting for publishing thread to stop....");
     stop();
     while (is_running()) {
       std::this_thread::sleep_for(std::chrono::microseconds(100));
     }
+    RCLCPP_DEBUG_STREAM(
+      rclcpp::get_logger("realtime_tools"), "Publishing thread stopped, joining thread....");
     if (thread_.joinable()) {
       thread_.join();
     }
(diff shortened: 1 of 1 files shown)


realtime_tools/include/realtime_tools/realtime_publisher.hpp
```diff
@@ -117,7 +121,10 @@ class RealtimePublisher
    */
   void stop()
   {
-    keep_running_ = false;
+    {
+      std::unique_lock<std::mutex> lock(msg_mutex_);
+      keep_running_ = false;
```
Juliaj, line 126: For the vulnerability you mentioned,  is it a potential race condition where `is_running_` might not be set as expected? 

https://github.com/ros-controls/realtime_tools/blob/51163648496ac539f2cee6811efd1f01141b2854/realtime_tools/include/realtime_tools/realtime_publisher.hpp#L241

saikishor: No, we are doing lock and unlock manually. If the code after lock fails, then the thread will be left dangling

saikishor: I'll open a PR tonight or tomorrow and I'll ask for your review @Juliaj

Juliaj (commented):
The change looks good to me. Could we add a test to ensure the two flags `keep-running_` and `is_running_` are  set correctly, just to cover our bases?

saikishor:
> The change looks good to me. Could we add a test to ensure the two flags `keep-running_` and `is_running_` are set correctly, just to cover our bases?

@Juliaj how do we ensure that? The change happens in the destructor right?

Juliaj:
> how do we ensure that? The change happens in the destructor right?

Yeah, it would be difficult to check these flags once the object is destroyed. What about a test case that:
sets up the publisher, publishes some messages, calls the stop() method to terminate the publisher loop, then verifies the expected behavior after termination ?

This can be a brand-new test or an add-on to the existing test case `construct_destruct`.

saikishor:
> > how do we ensure that? The change happens in the destructor right?
> 
> Yeah, it would be difficult to check these flags once the object is destroyed. What about a test case that:
> sets up the publisher, publishes some messages, calls the stop() method to terminate the publisher loop, then verifies the expected behavior after termination ?
> 
> This can be a brand-new test or an add-on to the existing test case `construct_destruct`. On a second thought, would `construct_destruct` cover this ?

Yes it would already be in construction and destruction. Let's wait for the new PR I'm making

christophfroehlich:
fixes most likely #307 also?

realtime_tools/include/realtime_tools/realtime_publisher.hpp
```diff
@@ -99,10 +99,14 @@ class RealtimePublisher
   /// Destructor
   ~RealtimePublisher()
   {
+    RCLCPP_DEBUG_STREAM(
```
christophfroehlich, line 102: why using STREAM here?

christophfroehlich (dismissed):
At he first look I don't see why the mutex should solve the problem, but I believe you ;)

realtime_tools/include/realtime_tools/realtime_publisher.hpp
saikishor: I can fix it

saikishor: I can fix it

saikishor: Done!

saikishor:
> At he first look I don't see why the mutex should solve the problem, but I believe you ;)

When we use notify_one, if the conditional variable is notified much earlier than it is waiting, then we will have the deadlock here, so I'm making sure that I get the mutex, such a way I update the value before

christophfroehlich (approved):
gotcha

PR #331, opened 2025-05-21
Use target_link_libraries instead of ament_target_dependencies

saikishor:
Fixes recent breaking change in rolling

### DIFF
--- a/realtime_tools/CMakeLists.txt
+++ b/realtime_tools/CMakeLists.txt
@@ -96,7 +96,7 @@ if(BUILD_TESTING)
   target_link_libraries(realtime_server_goal_handle_tests realtime_tools ${test_msgs_TARGETS})
 
   ament_add_gmock(test_async_function_handler test/test_async_function_handler.cpp)
-  target_link_libraries(test_async_function_handler PUBLIC
+  target_link_libraries(test_async_function_handler
                         realtime_tools
                         thread_priority
                         ${lifecycle_msgs_TARGETS}
--- a/realtime_tools/CMakeLists.txt
+++ b/realtime_tools/CMakeLists.txt
@@ -96,7 +96,11 @@ if(BUILD_TESTING)
   target_link_libraries(realtime_server_goal_handle_tests realtime_tools ${test_msgs_TARGETS})
 
   ament_add_gmock(test_async_function_handler test/test_async_function_handler.cpp)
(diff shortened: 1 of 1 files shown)


realtime_tools/CMakeLists.txt
```diff
@@ -80,29 +80,27 @@ if(BUILD_TESTING)
...
                   test/realtime_server_goal_handle.test
                   test/realtime_server_goal_handle_tests.cpp)
-  target_link_libraries(realtime_server_goal_handle_tests realtime_tools)
-  ament_target_dependencies(realtime_server_goal_handle_tests test_msgs)
+  target_link_libraries(realtime_server_goal_handle_tests realtime_tools ${test_msgs_TARGETS})
 
   ament_add_gmock(test_async_function_handler test/test_async_function_handler.cpp)
-  target_link_libraries(test_async_function_handler realtime_tools thread_priority)
-  ament_target_dependencies(test_async_function_handler lifecycle_msgs rclcpp_lifecycle)
+  target_link_libraries(test_async_function_handler PUBLIC
```
christophfroehlich, line 99:
```suggestion
  target_link_libraries(test_async_function_handler
```
maybe?

realtime_tools/CMakeLists.txt
```diff
@@ -80,29 +80,27 @@ if(BUILD_TESTING)
   target_link_libraries(realtime_buffer_tests realtime_tools)
 
   ament_add_gmock(lock_free_queue_tests test/lock_free_queue_tests.cpp)
-  if(WIN32)
-    # atomic is not found on Windows, but also not needed
-    target_link_libraries(lock_free_queue_tests realtime_tools Boost::boost)
-  else()
-    # without adding atomic, clang throws a linker error
-    target_link_libraries(lock_free_queue_tests realtime_tools atomic Boost::boost)
+  target_link_libraries(lock_free_queue_tests realtime_tools ${Boost_LIBRARIES})
```
fmauch, line 83: Is there a reason going for `${Boost_LIBRARIES}` instead of `Boost::boost`? The latter seems to be what modern CMake targets are supposed to look like.

fmauch (approved):
Looking good, feel free to ignore my comment, it's more out of interest.

realtime_tools/CMakeLists.txt
saikishor: I used `Boost::boost` at first and changed it later. I can modify it

PR #335, opened 2025-05-22
Remove boost dependency in realtime_tools library

saikishor:


bmagyar (approved):

bmagyar:
I think there's more stuff to remove w.r.t. boost, no?

realtime_tools/CMakeLists.txt
```diff
@@ -44,7 +44,7 @@ target_include_directories(realtime_tools PUBLIC
   $<BUILD_INTERFACE:${PROJECT_SOURCE_DIR}/include>
   $<INSTALL_INTERFACE:include/realtime_tools>
 )
-target_link_libraries(realtime_tools PUBLIC rclcpp::rclcpp rclcpp_action::rclcpp_action rcpputils::rcpputils Threads::Threads Boost::boost)
```
christophfroehlich, line 47: As by investigation of @saikishor, adding Boost::boost here does not link in the traditional sense but ensures that the include directories are setup

PR #371, opened 2025-07-26
Cleanup deprecations

christophfroehlich:
Once https://github.com/ros-controls/ros2_controllers/pull/1830 is merged.

Those have been out for at least one release cycle on rolling and [kilted](https://discourse.openrobotics.org/t/preparing-for-kilted-sync-and-patch-release-2025-07-25/49245).


saikishor (approved):

realtime_tools/include/realtime_tools/realtime_publisher.hpp
christophfroehlich, line 1: @saikishor is `get_msg()` threadsafe?

christophfroehlich: but should it be part of public Api at all?

saikishor: Maybe yes, if not we cannot use it in the tests right?

realtime_tools/include/realtime_tools/realtime_publisher.hpp
```diff
@@ -195,76 +179,6 @@ class RealtimePublisher
...
-    msg_mutex_.unlock();
-    updated_cond_.notify_one();
-  }
-
-  /**
-   * \brief Get the thread object for the publishing thread.
-   *
-   * This can be used to set thread properties.
-   */
   std::thread & get_thread() { return thread_; }
```
christophfroehlich, line 182:
```suggestion
  /**
   * \brief Get the thread object for the publishing thread.
   *
   * This can be used to set thread properties.
   */
  std::thread & get_thread() { return thread_; }
```

saikishor (approved):

PR #372, opened 2025-07-28
Only accept callable with T&

christophfroehlich:
Earlier, something like this was accepted
`rt_active_goal_.set([](auto goal) { goal = RealtimeGoalHandlePtr(); });`

Note the call-by-value, it won't change the internal value of the box, without any notice. Took me some time to debug the client code.


saikishor (approved):

saikishor:
Thanks for fixing it @christophfroehlich

PR #383, opened 2025-08-13
Add DETACHED scheduling policy for Async function handler

saikishor:
This PR is mostly to have a detached mode from the CM from triggering the cycles and this is needed for the hardware whose deadlines are pretty strict and any delay caused by update loop won't affect it


christophfroehlich (approved):
I don't look through every change, but the intention is clear and LGTM as you have tested it on your hardware.

PR #391, opened 2025-09-03
Fix the failing ament_cppcheck in realtime_thread_safe_box

saikishor:
Fixes:
```
ament_cppcheck...........................................................Failed
- hook id: ament_cppcheck
- exit code: 1

No problems found
No problems found
[realtime_tools/include/realtime_tools/realtime_thread_safe_box.hpp:227]: (error: missingReturn) Found a exit path from function with non-void return type that has missing return statement
1 errors
No problems found
No problems found

ament_cpplint............................................................Passed
```


christophfroehlich (approved):
LGTM

PR #402, opened 2025-09-12
Add AsyncFunctionHandlerParams initialization through ROS parameters

saikishor:
I've added a way to initialize the current parameter from the ROS nodes directly, much easier when using with nodes and controllers


christophfroehlich (commented):
in principle LGTM.

Do you plan to update https://github.com/ros-controls/ros2_control/pull/2477 with this?

saikishor:
> in principle LGTM.
> 
> Do you plan to update https://github.com/ros-controls/ros2_control/pull/2477 with this?

Not really 
I've just added it for future purposes. Maybe for the controller interface class

PR #355, opened 2025-06-11
Silence some warnings

christophfroehlich:
Since #323 we get lots of deprecation warnings. Some could be silenced, but the constructor+destructor still complains.

@saikishor any idea to fix this? Do we really need the deprecation warning for the `msg_` member?
```
In file included from /workspaces/ros2_rolling_ws/src/realtime_tools/realtime_tools/test/realtime_publisher_tests.cpp:39:
/workspaces/ros2_rolling_ws/src/realtime_tools/realtime_tools/include/realtime_tools/realtime_publisher.hpp: In instantiation of ‘realtime_tools::RealtimePublisher<MessageT>::RealtimePublisher(PublisherSharedPtr) [with MessageT = test_msgs::msg::Strings_<std::allocator<void> >; PublisherSharedPtr = std::shared_ptr<rclcpp::Publisher<test_msgs::msg::Strings_<std::allocator<void> >, std::allocator<void> > >]’:
/workspaces/ros2_rolling_ws/src/realtime_tools/realtime_tools/test/realtime_publisher_tests.cpp:71:42:   required from here
/workspaces/ros2_rolling_ws/src/realtime_tools/realtime_tools/include/realtime_tools/realtime_publisher.hpp:80:98: warning: ‘realtime_tools::RealtimePublisher<test_msgs::msg::Strings_<std::allocator<void> > >::msg_’ is deprecated: This variable is deprecated, it is recommended to use the try_publish() method instead. [-Wdeprecated-declarations]
   80 |   : publisher_(publisher), is_running_(false), keep_running_(true), turn_(State::LOOP_NOT_STARTED)
      |                                                                                                  ^
/workspaces/ros2_rolling_ws/src/realtime_tools/realtime_tools/include/realtime_tools/realtime_publisher.hpp:68:12: note: declared here
   68 |   MessageT msg_;
      |            ^~~~
```

PR #421, opened 2025-10-04
Deprecate `msg_` field

christophfroehlich:
I found a way in deprecating it without annoying deprecation warnings of the class itself.
Partially reverts #355 

Throws deprecation notices with current ros2_controller, but does not after
https://github.com/ros-controls/ros2_controllers/pull/1947


christophfroehlich:
what you mean? new syntax is only the windows part. what I meant what is new new here is disabling the deprecation warnigs for constructor+destructor, then it does not complain for the member variable

PR #423, opened 2025-10-05
Mark `msg_` variable in RT publisher as private

christophfroehlich:


saikishor:
Just for future reference: shouldn't be backported

Issue #279, opened 2025-02-10
Roadmap and guidelines for using Box/Buffer/Queue


christophfroehlich:
We want to summarize the differences between RealtimeBox and RealtimeBuffer, and compare it with the new Lock-Free Queue proposal. 
In the end, we should have guidelines when to use which concept and if maybe any of them is obsolete.

Finally, we will apply the guidelines to ros2_controllers wherever necessary.

## RealtimeBox

Got refurbed with #139 #146

[Sai]
- It is more a thread-safe way to accessing data than RT safe.
- Using pointers with this doesn't make a lot of sense as the data protected by the mutex or the wrapper can be modified and accessed without any restrictions. IMO, it shouldn't be the case with the RealtimeBox

## RealtimeBuffer

[Sai] RealtimeBuffer is what we mostly want, but we have some cons in this:
- As we all agreed in last WG meetings, NON_POLLING should be the default one now and it should work as expected https://github.com/ros-controls/realtime_tools/issues/212
- The member variables are using the pointer to the original data (We have to check what happens if the data goes out of scope and we try to access it)
- When accessing using the readFromRT etc, we get a raw pointer, may be this is not so nice.
- The other limitation I see is that the Buffer size is limited to size of 2.

[Denis] We also need “WriteFromRT”
  - https://github.com/ros-controls/ros2_controllers/issues/168 
  - https://github.com/ros-controls/realtime_tools/pull/73 

## Lock-Free Queue

[Sai] Was suggested in  https://github.com/ros-controls/realtime_tools/issues/14 and implemented with https://github.com/ros-controls/realtime_tools/pull/253 

- We don't use mutexes for the implementation
- We can have the buffer size as we need
- Can work with multi producer and multi consumer and single producer and single consumer efficiently
- and when we pop, we directly get the data instead of the pointer to the object

Example usage: https://github.com/ros-controls/ros2_controllers/pull/1480

christophfroehlich:
👀 Could you have a look please, and add your thoughts? And feel free to ping other developers if you think they can provide valuable input to this topic.
@destogl @bmagyar @firesurfer @fmauch @urfeex @saikishor @RobertWilbrandt

christophfroehlich:
Summary from WG meeting on February 12th:

- Rename RealtimeBox to RealtimeThreadsafeBox and use RealtimeMutexes @saikishor 
[...]

PR #447, opened 2025-11-03
Update RealtimeBuffer To Use Priority Mutex

bjin2364:
# Background
`RealtimeBuffers` is a bit of a misnomer, it's currently acting more like a thread-safe data structure. We can give it some more "real-time" elements using the mutex library in this package.

# Changes
- templatize `RealtimeBuffers` so users can provide their own mutex choice
  - default is the priority inherited mutex from [mutex.hpp](https://github.com/ros-controls/realtime_tools/blob/master/realtime_tools/include/realtime_tools/mutex.hpp)
- Update unit tests
- Switched `realtime_buffers_test.cpp` to use `gtest`. It doesn't look like there are `gmock` features used in the test.
- Added `ament_cmake_gtest` as a dependency... I think this may be a missing dep in general.

# Tests
- unit tests

### DIFF
--- a/realtime_tools/include/realtime_tools/realtime_buffer.hpp
+++ b/realtime_tools/include/realtime_tools/realtime_buffer.hpp
@@ -41,7 +41,7 @@
 
 namespace realtime_tools
 {
-  template <class T, class MutexType=prio_inherit_mutex>
+template <class T, class MutexType = prio_inherit_mutex>
 class RealtimeBuffer
 {
 public:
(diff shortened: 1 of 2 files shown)


christophfroehlich:
RealtimeBuffer is rather old and will possibly be deprecated and removed in the future. 
We had a long discussion about this, e.g. #279, resulting in the guidelines published here: https://control.ros.org/rolling/doc/realtime_tools/doc/index.html

saikishor (commented):
@bjin2364 Maybe the RealtimeThreadSafeBox is right to use for this case right?
 https://github.com/ros-controls/realtime_tools/blob/master/realtime_tools/include/realtime_tools/realtime_thread_safe_box.hpp

bjin2364:
ah TIL. I didn't take a close look at `realtime_thread_safe_box.hpp`, but it does accomplish what I was looking.

Will close this PR.

PR #454, opened 2025-11-13
Expand RealtimeBoxBestEffort

bjin2364:
# Changes
- Removes the static assertion so that `RealtimeBoxBestEffort` can be used with priority mutexes in `mutex.hpp`

# Tests
- type parameterized unit tests

### DIFF
--- a/ros_controls.kilted.repos
+++ b/ros_controls.kilted.repos
@@ -18,4 +18,4 @@ repositories:
   ros-controls/ros2_controllers:
     type: git
     url: https://github.com/ros-controls/ros2_controllers.git
-    version: master
+    version: kilted


saikishor (dismissed):
Looking pretty good to me 👌🏽👌🏽

bjin2364:
@christophfroehlich I don't think we need to cherry-pick to `master` branch because `master` already supports this: https://github.com/ros-controls/realtime_tools/blob/master/realtime_tools/include/realtime_tools/realtime_thread_safe_box.hpp#L69

christophfroehlich (approved):
Ok, I haven't looked into details before, the code diverged quite a lot from humble to newer distros. I'm happy to merge this into humble, could you still have a look if the improvements of the tests can be applied to the master branch? Thanks!

PR #459, opened 2025-11-23
Update ThreadSafeBox Tests

bjin2364:
# Changes

- Following up from https://github.com/ros-controls/realtime_tools/pull/454#pullrequestreview-3470451614, this PR improves test coverage by parameterizing the types in `realtime_thread_safe_box_tests` (similar test is in `humble`)

PR #460, opened 2025-11-23
Cleanup gmock Test Dependencies

bjin2364:
# Changes

- something that's slightly been bothering me are the `gmock` test deps, when none of the unit tests are using `gmock`. This PR updates the dependencies so they're `gtest`

# Testing

- passing unit tests

# Notes

- This PR is stacked on top of https://github.com/ros-controls/realtime_tools/pull/459


christophfroehlich:
we use gmock per default in all our repositories. I haven't compared it, but the argument of @bmagyar is that the failure messages are more descriptive.

PR #469, opened 2025-12-03
Create Publisher in RealtimePublisher

### LINKED ISSUE #81: realtime_tools API restructure

mamueluth:
This PR aims at creating the publisher inside the RealtimePublisher simplifying the creation of RealtimePublisher.

I chose to just add another constructor where you can pass the node and topic as well as qos on which we want to publish. I left the old constructor to give the user the flexibility to create or pass their own publisher if they want to.

If this approach is thought to simple let me know and i can change the implementation in a more desired direction.

### DIFF
--- a/realtime_tools/include/realtime_tools/realtime_publisher.hpp
+++ b/realtime_tools/include/realtime_tools/realtime_publisher.hpp
@@ -47,6 +47,7 @@
 #include <utility>
 
 #include "rclcpp/publisher.hpp"
+#include "rclcpp/node.hpp"
 
 namespace realtime_tools
 {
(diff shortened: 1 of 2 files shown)


christophfroehlich (commented):
Have you installed pre-commit? ;)

What would be a usecase for still passing a publisher to the constructor, and not deprecating the old API?

PR #473, opened 2025-12-23
Feature "async" - "slave" execution of HW Components enabling synchronizaiton to a robot controller and therefore better communicaiton stability.

destogl:
We need something that enables us to synchronize the HW loops with the robot. Many robot RT interfaces have “blocking read” to send data when they expect new data. This is an approach to tackle this.

> [!NOTE]
> will be rebased after on `master` after it is finished the testing on HW is successful.

### DIFF
--- a/realtime_tools/doc/async_scheduling_policy.rst
+++ b/realtime_tools/doc/async_scheduling_policy.rst
@@ -28,7 +28,7 @@ SLAVE
 -----
 In ``SLAVE`` mode, the async worker thread runs independently of the main thread, but *without any software rate-limiting or sleeping*. The handler continuously loops and immediately restarts the callback, expecting the callback function itself to block execution.
 
-Used when ehardware interfaces that must synchronize directly with an external hardware clock. Avoids drift between control loops of manipulator controllers and ``controller_manager`` thread. This requires the hardware interface to wait on an UDP heartbeat/sync signal from hardware that a new control cycle can start.
+Used when hardware interfaces must synchronize directly with an external hardware clock. Avoids drift between control loops of manipulator controllers and ``controller_manager`` thread. This requires the hardware interface to wait on a heartbeat/sync signal from hardware that a new control cycle can start.
 
 * **Scheduling Control:** Blocking in hardware interface ``read()`` function.
 * **Sleep Mechanism:** None. Relies on a blocking hardware ``read()`` to pace the thread.
--- a/realtime_tools/doc/async_scheduling_policy.rst
+++ b/realtime_tools/doc/async_scheduling_policy.rst
@@ -1,0 +1,34 @@
+Async Scheduling Policies
+=========================
+``realtime_tools`` allows hardware interfaces to be run in a thread separate from the main ``controller_manager`` thread.
+
+The ``AsyncFunctionHandler`` supports three distinct scheduling policies as to how the asynchronous worker thread executes its callback.
+
+.. _scheduling-policies:
+
+SYNCHRONIZED
+------------
+In ``SYNCHRONIZED`` mode, the async worker thread does not govern its own timing. Instead, it waits on a condition variable and is explicitly triggered by the main thread.
+
+Used when the asynchronous task must run in lockstep with the main control loop, but needs to be offloaded to a separate thread to prevent blocking the real-time path. Used with slow hardware interfaces doing heavy processing.
+
+* **Scheduling Control:** ``controller_manager`` main thread.
+* **Sleep Mechanism:** Waits on a ``contition_variable`` from the main thread.
+
+DETACHED
+--------
+In ``DETACHED`` mode, the async worker thread runs completely independently of the main ``controller_manager`` thread clock. It acts similarly to a standalone ROS 2 node, maintaining its own execution cycle based on a target ``exec_rate``.
+
+Hardware interfaces that need to poll or update at a fixed, independent frequency.
+
+* **Scheduling Control:** Internal software clock.
+* **Sleep Mechanism:** Calculates the elapsed time and explicitly calls ``std::this_thread::sleep_until()``, same as ``ros2_control_node`` executable.
+
+SLAVE
+-----
+In ``SLAVE`` mode, the async worker thread runs independently of the main thread, but *without any software rate-limiting or sleeping*. The handler continuously loops and immediately restarts the callback, expecting the callback function itself to block execution.
+
+Used when ehardware interfaces that must synchronize directly with an external hardware clock. Avoids drift between control loops of manipulator controllers and ``controller_manager`` thread. This requires the hardware interface to wait on an UDP heartbeat/sync signal from hardware that a new control cycle can start.
+
+* **Scheduling Control:** Blocking in hardware interface ``read()`` function.
+* **Sleep Mechanism:** None. Relies on a blocking hardware ``read()`` to pace the thread.
--- a/realtime_tools/include/realtime_tools/async_function_handler.hpp
+++ b/realtime_tools/include/realtime_tools/async_function_handler.hpp
@@ -90,6 +90,8 @@ public:
         return "synchronized";
       case DETACHED:
         return "detached";
+      case SLAVE:
+        return "slave";
       default:
         return "unknown";
     }
@@ -577,8 +579,10 @@ public:
         }
         if (params_.scheduling_policy == AsyncSchedulingPolicy::SYNCHRONIZED) {
           execute_synchronized_callback();
-        } else {
+        } else if (params_.scheduling_policy == AsyncSchedulingPolicy::DETACHED) {
           execute_detached_callback();
+        } else if (params_.scheduling_policy == AsyncSchedulingPolicy::SLAVE) {
+          execute_slave_callback();
         }
       });
     }
@@ -603,6 +607,54 @@ private:
           const auto end_time = std::chrono::steady_clock::now();
           last_execution_time_ =
             std::chrono::duration_cast<std::chrono::nanoseconds>(end_time - start_time);
+        }
+        trigger_in_progress_ = false;
+      }
+      cycle_end_condition_.notify_all();
+    }
+  }
+
+  void execute_slave_callback()
+  {
+    if (params_.exec_rate == 0u) {
+      throw std::runtime_error(
+        "AsyncFunctionHandler: Target Execution rate must be set when using SLAVE scheduling policy. It is used for cycle statistics reporting.");
+    }
+
+    if (pause_thread_) {
+      std::unique_lock<std::mutex> lock(async_mtx_);
+      async_callback_condition_.wait(
+        lock, [this] { return !pause_thread_ || stop_async_callback_; });
+    }
+
+    previous_time_ = params_.clock->now();
+
+    while (!stop_async_callback_.load(std::memory_order_relaxed)) {
+      {
+        // Handle pausing logic (e.g. during deactivation)
+        std::unique_lock<std::mutex> lock(async_mtx_);
+        if (pause_thread_) {
+             async_callback_condition_.wait(
+             lock, [this] { return !pause_thread_ || stop_async_callback_; });
+        }
+        
+        if (!stop_async_callback_) {
+          auto const current_time = params_.clock->now();
+          auto const measured_period = current_time - previous_time_;
+          previous_time_ = current_time;
+          current_callback_time_ = current_time;
+          current_callback_period_ = measured_period;
+
+          const auto exec_start = std::chrono::steady_clock::now();
+          try {
+            async_callback_return_ = async_function_(current_time, measured_period);
+          } catch (...) {
+            async_exception_ptr_ = std::current_exception();
+          }
+          last_execution_time_ = std::chrono::duration_cast<std::chrono::nanoseconds>(
+            std::chrono::steady_clock::now() - exec_start);
+
+          // No sleep! The loop immediately restarts, expecting the next async_function_ call to block.
         }
         trigger_in_progress_ = false;
       }
--- a/realtime_tools/include/realtime_tools/async_function_handler.hpp
+++ b/realtime_tools/include/realtime_tools/async_function_handler.hpp
@@ -44,6 +44,8 @@ namespace realtime_tools
  * thread will be triggering the async callback method.
  * DETACHED: The async worker thread will be detached from the main thread and will have its own
  * execution cycle.
+ * SLAVE: The async worker thread will be a slave to the hardware execution and will be blocked
+ * by it - usually when waiting for data in the read() method.
  * UNKNOWN: The scheduling policy is unknown.
  */
 class AsyncSchedulingPolicy
@@ -53,6 +55,7 @@ public:
     UNKNOWN = -1,  /// Unknown scheduling policy
     SYNCHRONIZED,  /// Synchronized scheduling policy
     DETACHED,      /// Detached scheduling policy
+    SLAVE,         /// Slave scheduling policy
   };
 
   AsyncSchedulingPolicy() = default;
@@ -63,6 +66,8 @@ public:
       value_ = SYNCHRONIZED;
     } else if (data_type == "detached") {
       value_ = DETACHED;
+    } else if (data_type == "slave") {
+      value_ = SLAVE;
     } else {
       value_ = UNKNOWN;
     }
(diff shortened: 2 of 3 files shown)


Nibanovic:
# Open questions

## how do controller `update()` loops fit in here?

For example, If we're running `update()` for a JTC that is targeting `async_slave` franka and we have another robot running in sync with controller manager, the `update()` of the controller is on a different clock than the `read/write`.

What are the knock-on effects of this and how do we address it? How do we establish a test case where the problems the controller causes are clear? **this is what I'm working on currently**

What first comes to mind is to run a controller as async controller. Then we configure a controller to target a specific hardware interface. Then, from the hardware interface, at the end of the `read()` we signal to the async controller to execute its `update()`. This would be the simplest approach to implement.

In the end, this could cause a mess of threads for a bunch of controllers, so maybe it would be useful to re-frame the async approach from "we spawn a thread for each async component" to "we spawn a thread for each **clock source that exists in our system**. 

Then, we'd then pin controllers `update` to be executed after a specific "clock source group" of hardware. This is a more extensive change, not so sure about it, but just and idea.

## phase-matching between ROS side and robot side, like in `kuka_rsi_driver`

Even if we solve the above mentioned problem, we still cannot handle controllers that target multiple hardware running on different clocks.

For this, we could take the existing approach in `kuka_rsi_driver` [for phase matching](https://github.com/fzi-forschungszentrum-informatik/kuka_rsi_driver/blob/main/kuka_rsi_driver/src/rsi.cpp#L207). This is more extensive and deserves a separate PR.

# Housekeeping
- overview overrun warnings. we probably want to suppress them in `cm` or in `async_function_handler`, as the robot clock we receive is subject to network jitter, depending on the network quality. This might also affect the `total_time` of execution we measure, as `read()` is basically blocking for the remainder of the total cycle time.
- Currently we still need **expected** update_rate defined for async slave hardware interfaces, even though `cm` is not scheduling it, for statistics calculations. This seems acceptable.

saikishor (changes requested):
I would recommend to go with the DETACHED methods as they have pretty much same code. Instead, you can add a parameter to the following `AsyncFunctionHandlerParams` like `skip_sleep` or `function_handle_rate` or something like that, so it is more maintainable IMO.

https://github.com/ros-controls/realtime_tools/blob/d666c2b1b44944184f7e34d072dd366d7e04148f/realtime_tools/include/realtime_tools/async_function_handler.hpp#L221-L229

destogl:
@urfeex this is ready :)

realtime_tools/doc/async_scheduling_policy.rst
```diff
@@ -0,0 +1,34 @@
...
+DETACHED
+--------
+In ``DETACHED`` mode, the async worker thread runs completely independently of the main ``controller_manager`` thread clock. It acts similarly to a standalone ROS 2 node, maintaining its own execution cycle based on a target ``exec_rate``.
+
+Hardware interfaces that need to poll or update at a fixed, independent frequency.
+
+* **Scheduling Control:** Internal software clock.
+* **Sleep Mechanism:** Calculates the elapsed time and explicitly calls ``std::this_thread::sleep_until()``, same as ``ros2_control_node`` executable.
+
+SLAVE
```
fmauch, line 27: I am a bit unsure whether the term "slave" is that catchy. At least my intuition is that there is a "master" thread probably running in the controller_manager that handles scheduling. I think, a more descriptive name like "hardware_sync" or "blocking_read" would make it more clear.

I am fine with "slave", though, as well. It might just be my intuition.

I like it very much that you highlighted the two aspects of scheduling control and sleep mechanism for each policy.

realtime_tools/doc/async_scheduling_policy.rst
```diff
@@ -0,0 +1,34 @@
...
+-----
+In ``SLAVE`` mode, the async worker thread runs independently of the main thread, but *without any software rate-limiting or sleeping*. The handler continuously loops and immediately restarts the callback, expecting the callback function itself to block execution.
+
+Used when ehardware interfaces that must synchronize directly with an external hardware clock. Avoids drift between control loops of manipulator controllers and ``controller_manager`` thread. This requires the hardware interface to wait on an UDP heartbeat/sync signal from hardware that a new control cycle can start.
```
fmauch:
```suggestion
Used when hardware interfaces must synchronize directly with an external hardware clock. Avoids drift between control loops of manipulator controllers and ``controller_manager`` thread. This requires the hardware interface to wait on a heartbeat/sync signal from hardware that a new control cycle can start.
```

realtime_tools/doc/async_scheduling_policy.rst
Nibanovic: Yep, this is a good question. I'm already used to "slave".  I'm mainly leaning towards it as it is descriptive (the loop is slave to a hardware clock), and the name is different enough to "stay" in your mind.

I'd love to hear others perspective on this.

saikishor (changes requested):
I'm skeptical about using mode called SLAVE. IMHO, the async function handler should be independent 

Given that it is doing exactly same as DETACHED mode, but just not maintaining any sleep cycle, whynot simply add a parameter called `maintain_cycle_rate` or something similar and then parse this from the other end.

Nibanovic:
> Given that it is doing exactly same as DETACHED mode, but just not maintaining any sleep cycle, whynot simply add a parameter called maintain_cycle_rate or something similar and then parse this from the other end.

My main point is that we need a named `AsyncSchedulingPolicy` for this feature, as when we use it in https://github.com/ros-controls/realtime_tools/pull/478 and https://github.com/ros-controls/ros2_control/pull/2971 to schedule controller updates in relation to blocking read, this enum is crucial.

I'd argue the users don't care if they are similar in implementation, they care about what it does for their hardware interface.

We can maybe rename SLAVE:
if SYNCHRONIZED is scheduled by the main thread, and DETACHED is scheduled by the hw-interface thread, we can maybe call it something like:
- EXTERNAL - scheduled by an external source
- BLOCKING - scheduled by the blocking read of the hardware interface
- HARDWARE_DRIVEN - self explanatory

destogl:
This is actually obsolete now.

Svastits:
@destogl why is this obsolate? it would be really helpful for KUKA multi-robot setups, as detached mode can sometimes break the timing

destogl:
> @destogl why is this obsolate? it would be really helpful for KUKA multi-robot setups, as detached mode can sometimes break the timing

We have decided in the Working group to not follow this approach, but a better one with the interface and not components synchronization. This is the approach from Orocos, and it should much more flexible than here proposed one.

On the other hand, this approach work. We are using it a lot and even with KUKA. Still we see it as intermediate solution for us until there is a “proper” one.

We will talk today in the WG meeting about this topic and see of other people are open to have this temporary solution merged, as it help already a lot.

destogl:
@Nibanovic let's just discuss the naming and see how to do merge this?

I propose "HW_SYNCHRONIZED"

saikishor:
> @Nibanovic let's just discuss the naming and see how to do merge this?
> 
> I propose "HW_SYNCHRONIZED"

Guys, what do you think of EXTERNALLY_PACED or EXTERNAL_CLOCK?

Nibanovic:
Im down for either `HARDWARE_DRIVEN` or `EXTERNALLY_PACED`. Leaning towards `HARDWARE_DRIVEN`, as it is, well, driven by the hardware.

`HW_SYNCHRONIZED` is not clear enough imo, and might be confused with `SYNCHRONIZED`

destogl:
@Nibanovic we do following:

Keep `DETACHED` option only and add "hardware_synchronized=true/false" option for the "slave" functionality.

christophfroehlich:
haven't we said we automatically choose that based on an inf or nan update rate, and only report the hardware_synchronized status to the hardware component? so not adding a new parameter

destogl:
> haven't we said we automatically choose that based on an inf or nan update rate, and only report the hardware_synchronized status to the hardware component? so not adding a new parameter

You are also right, might be simpler. @Nibanovic, can you check this, of course we also need the minimal sleep implementation to not do CPU burning (should be there).

Minimal execution time fix: https://github.com/b-robotized-forks/realtime_tools/tree/fix/minimal-exec-time-for-async-components

Nibanovic:
@christophfroehlich 

> haven't we said we automatically choose that based on an inf or nan update rate, and only report the hardware_synchronized status to the hardware component? so not adding a new parameter

We did, and I came away with a different conclusion to the discussion:

> I'd retain SLAVE mode as the main designator of this way of operating a hardware interface.
It is very clear for configuration that SLAVE is different than DETACHED, and asking of people to configure Detached + skip-sleep parameter seems very unintuitive.

In my experience with dozens of drivers and async hardware interfaces, we almost never use DETACHED mode. We use either SYNCHRONIZED (rarely) and the primary use for us is this HARDWARE_DRIVEN (SLAVE) mode.

From a user perspective, it seems weird to "hide" the most common and categorically different mode of operation. To configure this, user then needs to configure both DETACHED scheduling policy **and** put NaN in the update rate. This combination of configurations would then imply our HARDWARE_DRIVEN mode.

Discussed this with @destogl again, and we feel like giving this mode of operation a named scheduling policy is the clearest and simplest from users perspective.

christophfroehlich:
I don't have hard feelings here, I just summarized the discussion. Your reasoning make sense for me.. Is this then ready for review?

saikishor:
> From a user perspective, it seems weird to "hide" the most common and categorically different mode of operation. To configure this, user then needs to configure both DETACHED scheduling policy **and** put NaN in the update rate. This combination of configurations would then imply our HARDWARE_DRIVEN mode.
> 
> Discussed this with @destogl again, and we feel like giving this mode of operation a named scheduling policy is the clearest and simplest from users perspective.

Sure we can do it as a separate mode, but when we set this mode, internally it can do that. However, if the naming can be something different from the current one, it is good for me. It's just this part of the library system is agnostic to hardware or any usage, that's why

Nibanovic:
> of course we also need the minimal sleep implementation to not do CPU burning (should be there).

For this, I've opted for using existing `exec_rate` instead of adding a new parameter.
If for some reason `read()` does not block, we detect that by seeing whether the cycle was shorter than `period/10`. Then, we sleep for the rest of the period, like it was `DETACHED` mode.

This also relates to following comment:
> haven't we said we automatically choose that based on an inf or nan update rate, and only report the hardware_synchronized status to the hardware component? so not adding a new parameter

I think that the `period` that comes from `exec_rate` is important to keep even in `HARDWARE_DRIVEN` mode, as it is used for [calculating periodicity in the controller_manager](https://github.com/ros-controls/ros2_control/blob/17e2c1a85452c94e9024c62ca097e37f3a27e2fc/controller_manager/src/controller_manager.cpp#L4675-L4676), and we can't just put it to `inf` or `NaN`.

That is why I kept the `exec_rate != 0` check even for `HARDWARE_DRIVEN` mode. It may not be a very strong reason, but I would not dig too much deeper here, due to the following conclusion from RT Workgroup earlier in the discussion:

> We have decided in the Working group to not follow this approach, but a better one with the interface and not components synchronization. This is the approach from Orocos, and it should much more flexible than here proposed one.
On the other hand, this approach work. We are using it a lot and even with KUKA. Still we see it as intermediate solution for us until there is a “proper” one.
We will talk today in the WG meeting about this topic and see of other people are open to have this temporary solution merged, as it help already a lot.

What i hear here is: "It is not a perfect or a pretty solution, but it gets the job done until we come around to making a better one with more extensive changes"

And reviewing all the impacts this PR from `realtime_tools` has on the rest of `ros2_control` qualifies as more extensive changes, imho.

Let me know what you think. Opening the PR as ready for review.

christophfroehlich (changes requested):
For me this is fine, thanks for updating the docs. 

There is a clang issue btw
```
  /home/runner/work/realtime_tools/realtime_tools/.work/target_ws/src/realtime_tools/realtime_tools/test/test_async_function_handler.cpp:542:52: error: implicit conversion changes signedness: 'unsigned int' to 'const int' [-Werror,-Wsign-conversion]
    542 |   const int expected_max_cycles = params.exec_rate + 200;
        |             ~~~~~~~~~~~~~~~~~~~   ~~~~~~~~~~~~~~~~~^~~~~

```

realtime_tools/doc/async_scheduling_policy.rst
```diff
@@ -0,0 +1,34 @@
...
+.. _scheduling-policies:
+
+SYNCHRONIZED
+------------
+In ``SYNCHRONIZED`` mode, the async worker thread does not govern its own timing. Instead, it waits on a condition variable and is explicitly triggered by the main thread.
+
+Used when the asynchronous task must run in lockstep with the main control loop, but needs to be offloaded to a separate thread to prevent blocking the real-time path. Used with slow hardware interfaces doing heavy processing.
+
+* **Scheduling Control:** ``controller_manager`` main thread.
+* **Sleep Mechanism:** Waits on a ``contition_variable`` from the main thread.
```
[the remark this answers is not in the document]
destogl:
```suggestion
* **Sleep Mechanism:** Waits on a ``condition_variable`` from the main thread.
```

realtime_tools/include/realtime_tools/async_function_handler.hpp
```diff
@@ -589,10 +605,20 @@ class AsyncFunctionHandler
...
-        } else {
-          execute_detached_callback();
+
+        switch (params_.scheduling_policy) {
+          case AsyncSchedulingPolicy::SYNCHRONIZED:
+            execute_synchronized_callback();
+            break;
+          case AsyncSchedulingPolicy::DETACHED:
+          case AsyncSchedulingPolicy::HARDWARE_DRIVEN:
+            execute_detached_callback();
```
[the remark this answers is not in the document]
destogl: This is out of scope here for now. As it is complex to solve and not probable.

realtime_tools/include/realtime_tools/async_function_handler.hpp
```diff
@@ -668,23 +699,33 @@ class AsyncFunctionHandler
...
+                  params_.exec_rate, time_diff + cm_period, overrun_count + 1);
+              }
+              next_iteration_time += (overrun_count * period);
+            }
+            std::this_thread::sleep_until(next_iteration_time);
+          } else if (params_.scheduling_policy == AsyncSchedulingPolicy::HARDWARE_DRIVEN) {
+            // Safety net: If the hardware interface does not block and returned instantly,
+            // fallback to 10% of the nominal execution rate period to prevent CPU burning.
+            const auto last_period = last_execution_time_.load(std::memory_order_relaxed);
+            const auto minimum_expected_period = period / 10;
```
fmauch, line 724: I'm just wondering whether this should be a parameter or not. But we can very well leave things as they are now and if it turns out that users would like to have this configurable, we can still change it.

realtime_tools/doc/async_scheduling_policy.rst
```diff
@@ -0,0 +1,34 @@
+Async Scheduling Policies
+=========================
+``realtime_tools`` allows hardware interfaces to be run in a thread separate from the main ``controller_manager`` thread.
```
fmauch, line 3: I don't know whether this would be the right place to address this, but I think in the context of asynchronous hardware interfaces it might be important to explain how controllers are synchronized to that. Especially with the policy added in this PR: From my point of view the main motivation of having a hardware-driven pacing is to avoid clock drift, as this will ensure that in each hardware control cycle there will be a new controller command available. If only the hardware interface runs at the hardware's pace, but the controller doesn't, this might not resolve the issues that a user might want to solve by using this.

realtime_tools/include/realtime_tools/async_function_handler.hpp
```diff
@@ -140,20 +148,25 @@ struct AsyncFunctionHandlerParams
...
+        RCLCPP_ERROR(
+          logger, "Clock must be set when using DETACHED or HARDWARE_DRIVEN scheduling policy.");
         return false;
       }
       if (exec_rate == 0u) {
-        RCLCPP_ERROR(logger, "Execution rate must be set when using DETACHED scheduling policy.");
+        RCLCPP_ERROR(
+          logger,
+          "Execution rate must be set when using DETACHED or HARDWARE_DRIVEN scheduling policy.");
         return false;
```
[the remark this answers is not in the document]
fmauch: This is true, but not necessarily an issue with this PR.

fmauch (approved):
This looks clean and with a restricted scope. I've added two comments, that would both not necessarily be blockers for this PR, so approving.

destogl:
We have another fan for `hardware_driven`: https://github.com/kroshu/kuka_robot_descriptions/pull/284#issuecomment-5756764704

christophfroehlich (changes requested):
clang is still failing

realtime_tools/doc/async_scheduling_policy.rst
Nibanovic: I'll add a comment noting this. The solution for controllers is in a followup PR, which is closed for merge here, but we've successfully used it: https://github.com/ros-controls/realtime_tools/pull/478

realtime_tools/include/realtime_tools/async_function_handler.hpp
Nibanovic: If anyone desires to have this configurable, I'll add it. I'm having a hard time thinking of a case where someone would want to configure this, I consider this just a safety if someone starts a regular HW interface as HARDWARE_DRIVEN, so their PC does not sieze up.

realtime_tools/include/realtime_tools/async_function_handler.hpp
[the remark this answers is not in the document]
Nibanovic: I'll add it, as we are now using the execution rate to calculate `minimal_period` needed to prevent CPU burning

PR #489, opened 2026-02-10
Test downstream packages in downstream CI job

christophfroehlich:
#486 introduced an issue in ros2_controllers tests, let's not only build but test downstream packages.


saikishor (approved):

saikishor:
Thank you and sorry for introducing the bug 🙏🏽
I've an alternative solution now, I'll need to test on the robot to verify the improvement

PR #490, opened 2026-02-10
Use realtime mutex and also try-lock while setting feedback

saikishor:
This should fix the regression for now and hopefully the realtimeness too


christophfroehlich (changes requested):
We have to find a solution for windows

PR #513, opened 2026-04-11
Add configure_sched_rr() for SCHED_RR scheduling policy

### LINKED ISSUE #449: realtime_helpers.cpp has option to only used sched_fifo
Currently, `realtime_helpers.cpp` offers `configure_sched_fifo` to set thread priority. however, the function sets the policy to **SCHED_FIFO**. In a PREEMPT_RT kernel, where multiple threads could have RT priority, **SCHED_FIFO** threads may not rotate. A **SCHED_RR** policy may be better.
 
https://github.com/ros-controls/realtime_tools/blob/e668fb560a53548247d69443ac8bf752470d8a33/realtime_tools/src/realtime_helpers.cpp#L74-L99

I've started to use `realtime_helpers.cpp` in ros2control hardware interfaces and think it would be better if there was an option to opt for another policy i.e SCHED_RR.

I propose adding a new `configure_sched_rr` function that uses the **SCHED_RR** policy.

https://stackoverflow.com/questions/9392415/linux-sched-other-sched-fifo-and-sched-rr-differences
https://man7.org/linux/man-pages/man7/sched.7.html

nsk126:
## Description

Implements the feature mentioned in https://github.com/ros-controls/realtime_tools/issues/449 about using SCHED_RR as the thread policy. 

Fixes https://github.com/ros-controls/realtime_tools/issues/449

Users can now call `configure_sched_rr(priority)` to set the `SCHED_RR` scheduling policy on a thread.

Used Claude sonnet 4.5 to look up if Windows and Apple support SCHED_RR. Also used it for the commit message.


destogl (dismissed):
Thanks

christophfroehlich (approved):

PR #552, opened 2026-07-21
package.xml: correct license to a valid SPDX

vermaete:
This was causing a warning in Yocto/OpenEmbedded


christophfroehlich (approved):
Thank you!

PR #560, opened 2026-07-31
refactor: deprecate RealtimeBuffer and its headerfile

SouriRishik:
## Description

This PR deprecates the old `RealtimeBuffer` class and its header file `realtime_buffer.hpp` in favor of the modern `RealtimeThreadSafeBox` API.

Specifically, it completes the following:
1. Deprecates the `RealtimeBuffer` class itself using the C++14 standard `[[deprecated]]` attribute.
2. Deprecates the `realtime_buffer.hpp` header file by introducing compiler warning diagnostics (`#warning` for GCC/Clang and `#pragma message` for MSVC), which can be silenced by defining the `SILENCE_DEPRECATION_WARNINGS` macro.

## Is this user-facing behavior change?

No

## Did you use Generative AI?

No

## Additional Information

Verified locally that the pre-commit checks run and pass successfully:

```bash
pre-commit run --files realtime_tools/include/realtime_tools/realtime_buffer.hpp

### DIFF
--- a/realtime_tools/include/realtime_tools/realtime_buffer.hpp
+++ b/realtime_tools/include/realtime_tools/realtime_buffer.hpp
@@ -37,10 +37,19 @@
 #include <mutex>
 #include <thread>
 
+#ifndef SILENCE_DEPRECATION_WARNINGS
+#ifdef _MSC_VER
+#pragma message( \
+  "This header is obsolete, please use \"realtime_tools/realtime_thread_safe_box.hpp\" instead.")
+#else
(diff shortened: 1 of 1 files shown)


christophfroehlich (changes requested):
Thank you, but maybe better to link to the [guidelines](https://control.ros.org/rolling/doc/realtime_tools/doc/index.html#guidelines) instead. ThreadSafeBox is not always the best option.

christophfroehlich (approved):
Thank you. Let's wait for the PR in ros2_controllers before deprecating it

PR #573, opened 2026-08-05
feat: create publisher internally in RealtimePublisher

SouriRishik:
## Description

This PR refactors `RealtimePublisher` by adding a new template constructor that creates the underlying publisher internally. This simplifies using `RealtimePublisher` by removing the need for users to manually instantiate and pass in an `rclcpp::Publisher` object.

This completes and completes the work started in stale PR #469.

Fixes #81

## Is this user-facing behavior change?

Yes (adds a new constructor to the public API and deprecates the old constructor).

## Did you use Generative AI?

No

## Additional Information

Verified locally that the pre-commit checks run and pass successfully:

```bash
pre-commit run --all-files

PR #576, opened 2026-08-23
feat: use promise/future for RealtimePublisher thread startup synchronization

SouriRishik:
## Description

This PR refactors `RealtimePublisher`'s thread startup synchronization to use clean C++ standard `std::promise` and `std::future` instead of the original sleep-based spin loop.

Key changes:
1. Replaced the spin loop (`while` with `sleep_for`) with a clean `started_future.wait()` in the constructor.
2. In the thread `publishingLoop`, we set `turn_ = State::REALTIME` and fulfill the promise `started_promise.set_value()` to unblock the constructor, preventing any potential startup race condition where `can_publish()` could return false immediately after construction.
3. Consolidated the constructor initialization logic into a private helper function `initialize()` to keep the constructors DRY.

This resolves the follow-up review comments from PR #573.

No.

## Did you use Generative AI?

No.

## Additional Information

Verified locally that the pre-commit checks run and pass successfully:

```bash
pre-commit run --all-files

### DIFF
--- a/realtime_tools/include/realtime_tools/realtime_publisher.hpp
+++ b/realtime_tools/include/realtime_tools/realtime_publisher.hpp
@@ -40,6 +40,7 @@
 #include <atomic>
 #include <chrono>
 #include <condition_variable>
+#include <future>
 #include <memory>
 #include <mutex>
 #include <string>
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

**Review by christophfroehlich — APPROVED** [review]
Thanks!

File: realtime_tools/include/realtime_tools/realtime_publisher.hpp [hunk 1]
```diff
@@ -238,15 +236,16 @@ class RealtimePublisher
    *
    * The loop continues until keep_running_ is set to false.
    */
-  void publishingLoop()
+  void publishingLoop(std::promise<void> started_promise)
```
saikishor, line 239: Here, don't we need to parse by reference?. Just out of curiosity

SouriRishik: Since `started_promise` is a local variable on the stack of the initialization function, passing it by reference would create a dangling reference once the constructor finishes and the stack frame is destroyed. 

Passing it by value (which moves the promise) ensures its lifetime is safely managed inside the thread's own execution context.

PR #582, opened 2026-09-02
Install libboost-devel instead of boost with python support

christophfroehlich:
## Description

Jobs failed with 

```
Error:   × failed to solve requirements of environment 'default' for platform 'win-64'
  ├─▶   × failed to solve the environment

  ╰─▶ Cannot solve the request because of: boost * cannot be installed because there are no viable options:
      ├─ boost 1.85.0 would require
      │  └─ libboost-python-devel ==1.85.0 py312h7e22eef_4, which cannot be installed because there are no viable options:
      │     └─ libboost-python-devel 1.85.0 would require
      │        └─ python >=3.12,<3.13.0a0, for which no candidates were found.

```

no

no

### DIFF
--- a/.github/workflows/rolling-binary-build-win.yml
+++ b/.github/workflows/rolling-binary-build-win.yml
@@ -26,7 +26,7 @@ concurrency:
 
 jobs:
   binary-windows:
-    uses: ros-controls/ros2_control_ci/.github/workflows/reusable-ros-tooling-win-build.yml@master
+    uses: ros-controls/ros2_control_ci/.github/workflows/reusable-ros-tooling-win-build.yml@fix-win
     with:
       ros_distro: rolling
       pixi_dependencies: libboost-devel compilers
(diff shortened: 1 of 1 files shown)


saikishor (dismissed):

.github/workflows/rolling-binary-build-win.yml
```diff
@@ -26,7 +26,7 @@ concurrency:
 
 jobs:
   binary-windows:
-    uses: ros-controls/ros2_control_ci/.github/workflows/reusable-ros-tooling-win-build.yml@master
+    uses: ros-controls/ros2_control_ci/.github/workflows/reusable-ros-tooling-win-build.yml@fix-win
```
christophfroehlich:
```suggestion
    uses: ros-controls/ros2_control_ci/.github/workflows/reusable-ros-tooling-win-build.yml@master
```

PR #586, opened 2026-09-11
Add trySetFeedback for rt access

christophfroehlich:
## Description
https://github.com/ros-controls/ros2_controllers/issues/2554#issuecomment-5641515339

The failure was caused by reusing a preallocated action feedback message across the realtime controller and non-realtime publisher threads without synchronizing access to its contents. setFeedback() protected only the shared-pointer handoff, while JTC continued mutating the same message as runNonRealtime() published it, allowing clients to receive fields from different update cycles. This produced inconsistent desired and actual timestamps and exposed the flaky scaling assertion, particularly at a scaling factor of 2.0.

no

GPT 5.6


saikishor (approved):
LGTM

